PR #70, opened 2017-07-12
Mousewheel zoom doesn't work when mousing over a point 

biovisualize:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes mapd-immerse #1819

## :smoking: Smoke Test
- [x] Works in chrome
- [x] Works in firefox
- [x] Works in safari
- [x] Works in ie edge
- [x] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/chart.css
+++ b/dist/chart.css
@@ -478,7 +478,7 @@ body {
     border: 3px solid white;
     transition: border .25s;
     box-shadow: 0 1px 4px 0 rgba(0, 0, 0, 0.4);
-    pointer-events: auto; }
+    pointer-events: none; }
   .dc-chart .map-popup, .dc-chart .map-point-wrap {
     width: 100%;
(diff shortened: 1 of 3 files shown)


ayyrickay:
This does fix the issue, but we do run into some potentially confusing UX - namely, when we zoom in, the cursor "loses" the point, but keeps the pop up. See below:

popup zoom issue


@edmapd @duhoang Mru and I think this should be filed as an issue, but I imagine it's finicky work. How should we move forward?

biovisualize:
The clean fix would be to properly zoom around the cursor. But this commit fixes the zoom issue, which is pretty nasty. So one idea would be to remove the tooltip when zooming, then merge this fix and open another issue for zooming around the mouse. If @edmapd has an opinion, now is the time to tell it :)

PR #98, opened 2017-08-29
Raster Layer API as Vega-Lite Spec

### LINKED ISSUE #97: Point Raster Layer API as Vega-Lite Specification
```
{
  transform?: {
    sample?: true,
    limit?: number
  },
  encoding: "point",
  x: {
    type: "quantitative",
    field: "lon"
  },
  y: {
    type: "quantitative",
    field: "lat"
  },
  size: number | "auto" | {
    type: "quantitative",
    field: string,
    domain: [number, number],
    range: [number, number]
  },
  color: HexCode | {
    type: "density",
    range: Array<HexCode>
  } | {
    type: "quantitative" | "ordinal"
    field: string,
    domain: [number, number] | Array<string>,
    range: Array<HexCode>
  }
}
```

mrblueblue:
fixes #97 

## TODO
- [x] fix draw engine apply filters
- [x] sampling
- [x] cap/limit api
- [x] poly
- [x] declarative legends
- [ ] polish + testing

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -58,7 +58,7 @@
     "no-constant-condition": 2, // unnecessary; usually a typo.
     "no-continue": 1, // makes reasoning about loops more difficult.
     "no-control-regex": 2, // usually a typo.
-    "no-debugger": 1, // for debugging only; shouldn't be committed.
+    "no-debugger": 2, // for debugging only; shouldn't be committed.
     "no-delete-var": 2, // only properties should be deleted.
     "no-div-regex": 0, // regex are difficult enough; also operator-assignment disallows /= operator.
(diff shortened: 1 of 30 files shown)


nytai:
The popup color appears to be always defaulting to blue, both on poly and point.

nytai:
We should create a tag on master right before merging this, as it's a pretty significant breaking change.

nytai:
Some issues with legends:

1. When going from stacked legend (1 quant, 1 cat) to single legend the stacked one sticks around. 
2. Somehow the categorical legend is getting out of sync with the state. In single layer mode, it only seems to display the top 5, yet in stacked multi-mode it shows 7.

PR #100, opened 2017-09-01
[DO NOT MERGE] Chrisv/simple map api

biovisualize:
This is a PR just to show an experiment. The goal was to prove that we could improve mapd-charting API a lot by just copy-pasting all functions into the right modules, basically pivoting from a mixin pattern to a module pattern decoupling connection/data/query/rendering and eventually events and observer. A lot of details are hardcoded, but it's enough to validate the strategy. It is meant as a conversation to figure out when is the right time to clean-up the API if it becomes difficult to move on with deliverables.


mrblueblue:
I think this shows that re-implementing the raster chart might not be as hard as we think it is.

However, several things to keep in mind and worth investigating:

* Behavior related to popups and hovering behavior that we haven't looked at yet

*Requirements around auto sizing points based on zoom levels, sampling the data based on a formula, etc.

* Accounting for geo and non-geo related behavior

As a side note, I think we can actually take the declarative API one step further and get rid of the notion of a separate "layer" class. In the end, there should just be one raster instance to deal with and one spec to pass in.

The "layers-as-separate-classes" API seems to be an unnecessary complication over the very simple and intuitive "give-me-one-spec" API. 

The important question is then: are these complications actually necessary? What patterns existed in the code that complicated everything for us?

biovisualize:
Good points. I think what makes it complicated is the chart registry not being linked with the DOM and having crossfilter called from everywhere. Another assumption is that the problem we have with dc is to have to manage its state and internals from the outside instead of just passing a new spec and let it figure out the rest, like Vega is doing, and like I prototyped here: https://github.com/mapd/mapd-immerse/pull/3057/files#diff-aa29c996d9569733054c3f09f0d90b9aR86

For this exploration, the idea was that decoupling these modules would make it easier to find the cause o bugs, prevent side-effects and remove a big part of the code that is just dead weight meant to satisfy the mixins pattern.

biovisualize:
The text widget bypasses a whole area that I don't like in the code, being instantiated right at the dashboard level. I'm pretty sure I can do the same with row chart, even if I already bypass most of this code, to illustrate how simple it's supposed to be with a Vega-like API.

PR #114, opened 2017-10-25
[POINT LAYER] add ability to add mark symbols

mrblueblue:


nytai:
I'm getting an exception using shape `hexagon` I don't think that one is supported.

vastcharade:
use `hexagon-horiz` for horizontally-aligned hexagons or `hexagon-vert` for vertically-aligned hexagons.

PR #119, opened 2017-11-04
throw error if redrawAllAsync is called before renderAllAsync

### LINKED ISSUE #107: Throw an error if redraw is called before a render
## :beetle: If bug
  - [x] Operating System and version: MacOS 10.12.5 
  - [x] Browser and version: Chrome 61.0.3163.91
  - [x] Steps to reproduce (including dashboard link):
  - [x] Description of issue:

When i check Example1 in your examples the line chart shows a nice dot with a popup on mousemove. My line chart however gives the following error on mousemove: 

schermafbeelding 2017-09-21 om 12 02 31

my implementation is as follows:

```typescript
  public draw(days?: number) {

    this.crossfilter
      .groupAll()
      .reduceMulti(this.dateExpression)
      .valuesAsync(true).then((chartBounds: IMapdChartBound) => {

        // if there is no data to display for the given time range
        // chartBounds will return null values.
        // here we manually set the chartbounds for the given time range
        if (chartBounds.minimum === null && chartBounds.maximum === null) {
          switch (days) {
            case 1:
              chartBounds.minimum = subDays(new Date(), days);
              chartBounds.maximum = new Date();
              break;
            case 7:
              chartBounds.minimum = subDays(new Date(), days);
              chartBounds.maximum = new Date();
              break;
            case 30:
              chartBounds.minimum = subDays(new Date(), days);
              chartBounds.maximum = new Date();
              break;
            case 90:
              chartBounds.minimum = subDays(new Date(), days);
              chartBounds.maximum = new Date();
              break;
            case 365:
              chartBounds.minimum = subDays(new Date(), days);
              chartBounds.maximum = new Date();
              break;
          }
        }

        // return correct timebin value so graph doesn't show to much straight lines
        const getTimeBin = (days: number) => {
          const binMap = {
            1: 'hour',
            7: 'hour',
            30: 'day',
            90: 'day',
            365: 'week',
          };

          return binMap[days];
        };

          // draw transactions chart
          const transactionsRecentChart = (window as any).dc.lineChart('.transactions__recent')
            .width(this.parentWidth)
            .height(300)
            .colors((window as any).chart_colors)
            .transitionDuration(1500)
            .brushOn(true)
            .elasticY(true)
            .enablePopup(true)
            .dimension(this.transactionsRecent)
            .group(this.transactionsRecentGroup)
            .x(
              (window as any).d3.time.scale.utc()
                .domain([chartBounds.minimum, chartBounds.maximum]),
            )
            .binParams({
              binBounds: [chartBounds.minimum, chartBounds.maximum],
              timeBin: getTimeBin(days),
            });

          transactionsRecentChart
            .xAxis()
            .scale(transactionsRecentChart.x())
            .tickFormat((window as any).dc.utils.customTimeFormat);

        (window as any).dc.redrawAllAsync();
      });
``` 
This problem occurs both with the linechart and also with the piechart. Since my linechart implementation does not differ a lot from what you do in your example 1 i'm thinking i'm might missing a dependency. 

Can you point me in the right direction?

luukgruijs:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #107 

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.css
+++ b/dist/mapdc.css
@@ -368,93 +368,6 @@ art {
 .dc-chart .heat-box:hover{
     fill-opacity:0.8;
 }
-
-/* loading */
-
-/* spinner */
-
-.chart-loading-overlay {
-  opacity: 0.5;
-  pointer-events: none;
-}
-
-.loading-widget-dc {
-  font-size: 12px;
-  position: absolute;
-  top: calc(50% - 24px);
-  left: calc(50% - 24px);
-  transform: translate(-50%, -50%);
-  color: blue;
-  text-transform: uppercase;
-  text-align: center;
-  font-weight: bold;
-}
-.loading-widget-dc .loading-msg {
-  margin-top: 24px;
-}
-
-.waiting, .waiting *{
-  cursor: wait !important;
-}
-
-.main-loading-icon {
-  position: absolute;
-  width: 48px;
-  height: 48px;
-  border: 4px solid transparent;
-  border-top-color: blue;
-  border-radius: 50%;
-  animation: mainLoadIconAnimation .6s infinite linear;
-}
-
-.main-loading-icon:before, .main-loading-icon:after {
-  z-index: -1;
-  position: absolute;
-  content: "";
-  width: calc(100% + 8px);
-  height: calc(100% + 8px);
-  border-radius: 50%;
-  border: 4px solid transparent;
-  border-top-color: blue;
-  top: -4px;
-  left: -4px;
-  animation: beforeGrow .3s infinite linear alternate;
-}
-
-.main-loading-icon:after {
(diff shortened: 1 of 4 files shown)


luukgruijs:
this PR now throws an error if `redrawAllAsync()` is called before `renderAllAsync()`. Would it be preferable to apply this also to `redrawAsync()` and `renderAsync()`?

nytai:
> Would it be preferable to apply this also to redrawAsync() and renderAsync()?

@luukgruijs, sure that sounds like a great idea, it can be a separate commit though. This one can go as is. 

Thank you for the contribution!

PR #128, opened 2017-11-30
[SCATTER CHART] add legendables

nytai:


duhoang:
👎  It's been an incredible journey.

PR #129, opened 2017-12-05
Vega lite to support aggregations for point-layers

nytai:


micahstubbs:
looks like we have an approval from another dev.  is there anything else we need to merge this PR?

PR #131, opened 2017-12-08
Enable table text select without filter

clhenrick:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes https://github.com/mapd/mapd-immerse/issues/3583

## :smoking: Smoke Test
- [x] Works in chrome
- [x] Works in firefox
- [x] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/charts/mapd-table.js
+++ b/src/charts/mapd-table.js
@@ -337,6 +337,14 @@ export default function mapdTable (parent, chartGroup) {
                     })
                     .classed("active", _sortColumn ? _sortColumn.index === i : false)
                     .classed(_sortColumn ? _sortColumn.order : "", true)
+                    .style("width", d3.select(this).node().getBoundingClientRect().width + "px")
+
+
+              const textSpan = sortLabel.append("span")
+                    .text(d.label)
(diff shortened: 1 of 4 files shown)


src/charts/mapd-table.js
```diff
@@ -291,6 +291,10 @@ export default function mapdTable (parent, chartGroup) {
         .html(d => formatDataValue(d[col.name]))
         .classed("filtered", col.expression in _filteredColumns)
         .on("click", d => {
+          // detect if user is selecting text or clicking a value, if so don't filter data
+          const s = window.getSelection().toString()
+          if (s.length) { return }
```
nytai: I tend to favor condensing statements, so putting `window.getSelection().toString()` directly in the if condition, as more variables = more to remember. Just a personal preference though.

clhenrick: Makes sense. I considered making this a one liner, but thought that making it more verbose would be slightly more readable for future devs.

PR #142, opened 2018-01-18
micah/ff bbox bugfix

micahstubbs:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes https://github.com/mapd/mapd-immerse/issues/3697
- [x] Works in Firefox  
- [x] Works in Chrome  
- [x] Works in Safari  
- [x] Works in Edge  
- [x] Works in IE  

🎉  both layers render now:

Layer 1
screen shot 2018-01-18 at 3 31 30 pm

Layer 2
screen shot 2018-01-18 at 3 31 39 pm


src/charts/row-chart.js
```diff
@@ -411,8 +411,21 @@ export default function rowChart (parent, chartGroup) {
                   const thisLabel = d3.select(this)
 
                   const width = Math.abs(rootValue() - _x(_chart.valueAccessor()(d)))
-                  const measureWidth = thisLabel.node().getBBox().width
-                  const dimWidth = _chart.svg().select("text.value-dim._" + i).node().getBBox().width
+                  
+                  //
+                  // handle Firefox getBBox bug
```
nytai: This comment maybe isn't so helpful, it certainly won't be in a few months. Maybe link to the "bug"

micahstubbs: good call

PR #148, opened 2018-01-24
gather shapefilters in statement; apply statement to all layers

nytai:


alexbaden:
Can confirm this resolves the lasso / selection chart issue.

PR #169, opened 2018-02-26
Guard invalid continuous legend start and end values

jonvuri:
Fixes behavioral issue in https://github.com/mapd/mapd-immerse/issues/3942, by short-circuiting a change to the start value and end value inputs on a continuous legend if the start value is greater than the end value.


micahstubbs:
@jrajav approved, please merge once the CI builds are green

PR #188, opened 2018-04-19
Fix formatting in pointmap and scatterplot

biovisualize:
See https://github.com/mapd/mapd-immerse/issues/4390
# Merge Checklist
## :wrench: Issue(s) fixed:

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -52297,7 +52297,10 @@ function rasterLayer(layerType) {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


biovisualize:
The fix here was about the `key` that can sometimes be like `AVG(air_time)` and need to be extracted as `air_time` to be used as a proper key to match the formatters. That pattern happens elsewhere in the code too.

PR #192, opened 2018-04-24
lasso filter clearing bug from chart level on dashboard

uyanga-gb:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #https://github.com/mapd/mapd-immerse/issues/4286

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


src/mixins/raster-draw-mixin.js
```diff
@@ -464,6 +464,9 @@ export function rasterDrawMixin(chart) {
       }
       const shapes = drawEngine.sortedShapes
       drawEngine.deleteAllShapes()
+
+      origFilterFunc(Symbol.for("clear"))
```
biovisualize: I don't fully understand the fix but it makes sense with regards to how I see this used in other parts of the code. So we don't really need to worry that [Symbol.for](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for) is not available on IE, as we probably use a shim.

biovisualize:
Pointing to crossfilter latest master will probably fix Nigthwatch. [bf77150] I have a PR ready to do that https://github.com/mapd/mapd-charting/pull/193 If you apply it here instead, feel free to close it.

Issue #195, opened 2018-04-25
Revert workaround for mapd-immerse/issues/3880


steveblackmon-mapd:
## :beetle: If bug
  - [ ] Operating System and version:
  - [ ] Browser and version:
  - [ ] Steps to reproduce (including dashboard link):
  - [ ] Description of issue:

## :new: If feature request  
  - [ ] User story:
  - [ ] Specific requirements/functionality:
  - [ ] Mocks (if applicable):

This is effectively reverting a workaround for a back end rendering performance issue. The original bug is here:

https://github.com/mapd/mapd-immerse/issues/3880

We have since implemented a different method of rendering symbols which no longer suffers from this problem (it's very similar to the old 'points' mark but works for all symbol shapes). Backend rendering was switched over to the new rendering method at commit aef949dd70d4abac6790ba7adeebfa04404d9ad5.

steveblackmon-mapd:
Fixed at e677b03919368cc6ff54cd086935c70ca36a1201

PR #196, opened 2018-04-25
Remove check for 'circle' mark type and its replacement with points

steveblackmon-mapd:
This change should be invisible to users, but will allow us to fully deprecate the old point marks and always use circle symbols instead (which was the original intent).

It requires mapd-core to be beyond commit aef949dd70d4abac6790ba7adeebfa04404d9ad5

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Author referenced issue(s) fixed by this PR: mapd-charting #195
- [ ] Fixes mapd-charting and #195 mapd-immerse #3880

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/raster-layer-point-mixin.js
+++ b/src/mixins/raster-layer-point-mixin.js
@@ -459,42 +459,28 @@ export default function rasterLayerPointMixin(_layer) {
 
     const marks = [
       {
-        type: markType === "circle" ? "points" : "symbol",
+        type: "symbol",
         from: {
           data: layerName
         },
         properties: Object.assign(
           {},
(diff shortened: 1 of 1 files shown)


steveblackmon-mapd:
We've found a visual bug in the new symbol rendering method when rendering binned heat maps that I want to fix before I merge this PR.

steveblackmon-mapd:
Killing this PR as the affected files have diverged too much. Will open a new one.

PR #197, opened 2018-04-26
Fixing gradient legend style for dark maps

menaczar:
Fixes legend colors for dark maps:
### Before:
screen shot 2018-04-25 at 11 28 18 pm

### After: 
screen shot 2018-04-25 at 11 31 55 pm

### DIFF
--- a/scss/chart.scss
+++ b/scss/chart.scss
@@ -72,8 +72,9 @@ body {
 .basemap-satellite, .basemap-dark, .basemap-odyssey, .basemap-decimal {
     .gradient-legend .block .text  {
         color: #eeeeee;
+        background: rgba(0,0,0,0.1);
     }
-    
+
     .mapboxgl-ctrl-attrib {
         background: rgba(black, .4);
     }
--- a/scss/chart.scss
+++ b/scss/chart.scss
@@ -72,7 +72,8 @@ body {
 .basemap-satellite, .basemap-dark, .basemap-odyssey, .basemap-decimal {
     .gradient-legend .block .text  {
         color: #eeeeee;
-        background: rgba(0,0,0,0.1);
+        text-shadow: -1px -1px 0 #000, 1px -1px 0 #000, -1px 1px 0 #000, 1px 1px 0 #000;
+        background: rgba(0,0,0,0);
     }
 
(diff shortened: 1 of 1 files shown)


scss/chart.scss
```diff
@@ -70,6 +70,11 @@ body {
 }
 
 .basemap-satellite, .basemap-dark, .basemap-odyssey, .basemap-decimal {
+    .gradient-legend .block .text  {
+        color: #eeeeee;
+        background: rgba(0,0,0,0.1);
```
jrajav: Why even 0.1 alpha? Could this not just be transparent?

biovisualize:
Thanks @menaczar! Can you open a PR in Immerse that points package.json to this fix so we can test it? I'm curious to see how basemap-satellite, basemap-odyssey and basemap-decimal will look like with the nearly white font. 

Another trick could be to use a white font with an outline. A nice cross-browser css hack is
```
color: white;
text-shadow: -1px -1px 0 #000, 1px -1px 0 #000, -1px 1px 0 #000, 1px 1px 0 #000;
background: none;
```
On white background (new style on the right)

screen shot 2018-04-26 at 9 54 40 am

On black background

screen shot 2018-04-26 at 9 54 05 am

scss/chart.scss
menaczar: @jrajav I think this looks better on the light areas of the dark maps.
I'll open a PR in immerse to test thso

scss/chart.scss
```diff
@@ -70,6 +70,12 @@ body {
 }
 
 .basemap-satellite, .basemap-dark, .basemap-odyssey, .basemap-decimal {
+    .gradient-legend .block .text  {
+        color: #eeeeee;
+        text-shadow: -1px -1px 0 #000, 1px -1px 0 #000, -1px 1px 0 #000, 1px 1px 0 #000;
+        background: rgba(0,0,0,0);
```
biovisualize: The intention would be more clear if you would set `background: none`. `background` is a shorthand for setting all background properties, including `background-color`. So another alternative would be to use `background-color: transparent`.

menaczar:
@biovisualize its ready now :D

menaczar:
merging per ok from Chris V

PR #202, opened 2018-05-02
point map Other color bug fixed

uyanga-gb:
Relates to Immerse https://github.com/mapd/mapd-immerse/issues/4334
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -26020,7 +26020,7 @@ function getScales(_ref2, layerName) {
       range: color.range.map(function (c) {
         return (0, _utilsVega.adjustOpacity)(c, color.opacity);
       }),
-      default: (0, _utilsVega.adjustOpacity)("#27aeef", color.opacity),
+      default: (0, _utilsVega.adjustOpacity)(color.range[color.range.length - 1], color.opacity), // in current implementation 'Other' is always added as last element in the array
       nullValue: (0, _utilsVega.adjustOpacity)("#CACACA", color.opacity)
     });
   }
--- a/src/mixins/raster-layer-point-mixin.js
+++ b/src/mixins/raster-layer-point-mixin.js
@@ -247,7 +247,10 @@ function getScales({ size, color }, layerName) {
       type: "ordinal",
       domain: color.domain,
       range: color.range.map(c => adjustOpacity(c, color.opacity)),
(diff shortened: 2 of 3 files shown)


clhenrick:
@uyanga-gb looks like the CI builds are failing, might be because of a Prettier formatting issue.

uyanga-gb:
@clhenrick, yeah, thank you, It was unit test, and it is fixed now.

src/mixins/raster-layer-point-mixin.js
```diff
@@ -247,7 +247,10 @@ function getScales({ size, color }, layerName) {
       type: "ordinal",
       domain: color.domain,
       range: color.range.map(c => adjustOpacity(c, color.opacity)),
-      default: adjustOpacity("#27aeef", color.opacity),
```
uyanga-gb: Only this part is changed. Please ignore other prettier formatting.

dist/mapdc.js
```diff
@@ -26020,7 +26020,7 @@ function getScales(_ref2, layerName) {
       range: color.range.map(function (c) {
         return (0, _utilsVega.adjustOpacity)(c, color.opacity);
       }),
-      default: (0, _utilsVega.adjustOpacity)("#27aeef", color.opacity),
+      default: (0, _utilsVega.adjustOpacity)(color.range[color.range.length - 1], color.opacity), // in current implementation 'Other' is always added as last element in the array
```
biovisualize: This is better than an hardcoded value, but it will break if we change the logic that makes the "Other" color be the last of the color range. Still I suggest we keep it like this as it's simpler than other alternatives.

For example, just for the record, a more extensible way would be to repurpose the 
```
_chart.defaultColorAccessor = function() {
  return _defaultAccessor
}
```
from the color mixin that is not used anywhere else in the code (and even have a `// what is this?` comment right above). But it would involve bringing the color mixin to this raster mixin, which may just make it more confusing. So let's just leave it like this and revisit if needed.

PR #203, opened 2018-05-08
fixed point map initial LastFilteredSize bug

uyanga-gb:
Immerse PR https://github.com/mapd/mapd-immerse/pull/4464
Issue https://github.com/mapd/mapd-immerse/issues/4046

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/raster-layer-point-mixin.js
+++ b/src/mixins/raster-layer-point-mixin.js
@@ -247,7 +247,10 @@ function getScales({ size, color }, layerName) {
       type: "ordinal",
       domain: color.domain,
       range: color.range.map(c => adjustOpacity(c, color.opacity)),
-      default: adjustOpacity("#27aeef", color.opacity),
+      default: adjustOpacity(
+        color.range[color.range.length - 1],
+        color.opacity
+      ), // in current implementation 'Other' is always added as last element in the array
       nullValue: adjustOpacity("#CACACA", color.opacity)
     })
   }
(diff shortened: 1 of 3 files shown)


src/mixins/raster-layer-point-mixin.js
```diff
@@ -448,7 +448,44 @@ export default function rasterLayerPointMixin(_layer) {
...
+    _tot = number
+    return _layer
+  }
+
+  _layer.getTotalRecordsAsync = function(cf) {
+    if (_layer.tot()) {
+      return Promise.resolve()
+    }
+    return cf
+        .dimension()
```
jrajav: Wondering what the intent of this is - is it to just call this block of cf methods once and only once?

dist/mapdc.css
```diff
@@ -425,6 +425,14 @@ body {
   padding: 0 4px;
   text-decoration: none; }
 
+.basemap-satellite .gradient-legend .block .text, .basemap-dark .gradient-legend .block .text, .basemap-odyssey .gradient-legend .block .text, .basemap-decimal .gradient-legend .block .text {
+  color: #eeeeee;
```
jrajav: Let's try to be consistent with values - either `#eee` here or `#000000` below. (I'd prefer `#eee` for repeated values, more concise and still precise.)

This is a stylelint rule we could enable if we ran that.

Also, `#000` for `black` below.

dist/mapdc.css
```diff
@@ -425,6 +425,14 @@ body {
   padding: 0 4px;
   text-decoration: none; }
 
+.basemap-satellite .gradient-legend .block .text, .basemap-dark .gradient-legend .block .text, .basemap-odyssey .gradient-legend .block .text, .basemap-decimal .gradient-legend .block .text {
```
clhenrick: @uyanga-gb it might be good to dismiss these changes and ask @menaczar to add them in a separate PR as they don't relate to this PR. They are from: https://github.com/mapd/mapd-charting/pull/197/files

src/mixins/raster-layer-point-mixin.js
```diff
@@ -448,7 +448,44 @@ export default function rasterLayerPointMixin(_layer) {
...
+  }
+
   _layer._genVega = function(chart, layerName, group, query) {
+
+    // needed to set LastFilteredSize when point map first initialized
+    const cf = _layer.crossfilter()
+    if(!lastFilteredSize(cf.getId)){
+      if(cf.dimension){
+        _layer.getTotalRecordsAsync(cf).then(() => {
+          _layer.yDim().groupAll().valueAsync().then(value=>{
```
clhenrick: just a little nit-picky, watch the spacing here: `value=>{`, IMO it's more readable to have `value => {` which is more consistent with the rest of our libraries as well.

dist/mapdc.css
clhenrick: @jrajav these aren't changes from this PR, they are from https://github.com/mapd/mapd-charting/pull/197/files so I feel that we should include them in a separate PR

src/mixins/raster-layer-point-mixin.js
biovisualize: It's a copy-paste for a pattern found in the count widget. I agree that it could be made simpler and more clear, for exemple by using the `tot` variable directly, as it doesn't need to be a layer method.

dist/mapdc.css
uyanga-gb: @clhenrick, sure I will do that.

biovisualize: Here is the PR https://github.com/mapd/mapd-charting/pull/206
@uyanga-gb feel free to merge it

PR #208, opened 2018-05-11
column filter on measure will update global filter

uyanga-gb:
Description: previously the measure filter on column is only applied in the table chart not the global filter in dashboard level. Contrastingly, only dimension filter in table chart was modifying the global filter. However, the measure filter on column values were crossfiltering across other charts, so it was a confusing UX for the users if they are not sure dimension filter (grouped) is filtering differently than the measure filter. Thus, this change will apply the measure filter on column to the global filter as same as dimension filter. 
Check the build on https://github.com/mapd/mapd-immerse/pull/4484
Please check the description on the related immerse  PR to reproduce the reported issue.
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/filter-mixin.js
+++ b/src/mixins/filter-mixin.js
@@ -34,6 +34,9 @@ export function filterHandlerWithChartContext(_chart) {
 
     if (filters.length === 0) {
       dimension.filterAll(_chart.softFilterClear())
+      if(_chart.clearTableFilter){
+        _chart.clearTableFilter() // global filter also will clear all the columns filters on the table
+      }
     } else if (_chart.hasOwnProperty("rangeFocused")) {
       dimension.filterMulti(
         filters,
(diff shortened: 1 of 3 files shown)


src/mixins/filter-mixin.js
```diff
@@ -34,13 +34,18 @@ export function filterHandlerWithChartContext(_chart) {
...
+        _chart.clearTableFilter() // global filter also will clear all the columns filters on the table
+      }
     } else if (_chart.hasOwnProperty("rangeFocused")) {
       dimension.filterMulti(
         filters,
         _chart.rangeFocused(),
         _chart.filtersInverse(),
         _chart.group().binParams()
       )
+    } else if(_chart.getFilteredColumns && Object.keys(_chart.getFilteredColumns()).length>0){ // case for column filtering on measures
```
jrajav: Whitespace issues here - run eslint?

jrajav:
Other than the minor eslint issue, all fine - verified fix, and code looks good. Approving assuming we'll fix the eslint

uyanga-gb:
@jrajav, white space is fixed now.

PR #212, opened 2018-05-15
point webpack to compile.js

biovisualize:
Use a compile target for exposing dc as a global

### DIFF
--- a/compile.js
+++ b/compile.js
@@ -1,0 +1,1 @@
+require("expose-loader?dc!./src/compile.js")
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -7,7 +7,7 @@
 		exports["mapdc"] = factory(require("d3"));
 	else
 		root["mapdc"] = factory(root["d3"]);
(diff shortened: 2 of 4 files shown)


biovisualize:
I'm testing to see if it works. But I know That it compiles and sets a global dc object. It can be tested with a minimal html fixture.
```
<!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="Content-type" content="text/html; charset=utf-8"/>
    <script src="mapdc.js"></script>
  </head>
  <body>
    <div class="chart"></div>
    <script>
      console.log("mapd-charting", dc)
    </script>
  </body>
</html>
```

clhenrick:
I believe we still need to handle these lines from `src/index.js`:

```
require("../mapdc.css")
require("../scss/chart.scss")
require("mapbox-gl/dist/styles.css")
require("legendables/src/styles.css")

if (process.env.BABEL_ENV !== "test") {
  window.mapboxgl = require("mapbox-gl/dist/mapbox-gl.js")
  require("mapbox-gl/dist/mapboxgl-overrides.js")
}

require("./mixins/d3.box.js")

export * as d3 from "d3" // eslint-disable-line
```

PR #219, opened 2018-05-24
Uyanga/4502 ie chart label

uyanga-gb:
Description: Bar/row, Pie, and Bubble charts labels were not visible on IE.
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #https://github.com/mapd/mapd-immerse/issues/4502

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/charts/pie-chart.js
+++ b/src/charts/pie-chart.js
@@ -211,8 +211,8 @@ export default function pieChart(parent, chartGroup) {
         "deselected-label",
         d => _chart.hasFilter() && !isSelectedSlice(d)
       )
-      .html((d) => _chart.label()(d.data))
-      .html(function(d) {
+      .text((d) => _chart.label()(d.data))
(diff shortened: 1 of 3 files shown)


clhenrick:
@uyanga-gb so it seems D3's `.html()` doesn't play well with IE, but `.text()` does? Seems like there shouldn't be much difference between the two so I wonder if it's a IE bug in D3.

uyanga-gb:
@clhenrick, looks like html label in svg is treated as foreignObject, and that is not supported in IE. Text and rect are not using foreignObject, so that is probably why it is working with text. 
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject

PR #221, opened 2018-05-25
geo: stack mapd image layer between basemap labels & features

clhenrick:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] relates to https://github.com/mapd/mapd-immerse/issues/3784

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


src/mixins/map-mixin.js
```diff
@@ -521,6 +523,18 @@ export default function mapMixin(
...
+  function getFirstSymbolLayerId () {
+    let firstSymbolId = null
+    const layers = _map.getStyle().layers
+    for (let i = 0; i < layers.length; ++i) {
+      if (layers[i].type === "symbol") {
+        firstSymbolId = layers[i].id
+        break
+      }
+    }
+    return firstSymbolId
```
biovisualize: I guess nothing bad happens when firstSymbolId is null? I tried the various styles and it was all working.

dist/mapdc.js
```diff
@@ -43033,15 +43047,31 @@ exports.default = bbox;
 
 /***/ }),
 /* 196 */
-/***/ (function(module, exports, __webpack_require__) {
+/***/ (function(module, __webpack_exports__, __webpack_require__) {
```
biovisualize: You have the same compilation artifacts as me. I'm pretty sure we are doing something different than @uyanga-gb. It's not a problem, but it would make the diff cleaner to figure it out.

clhenrick: hmmm. I was running `npm run watch` which runs `webpack watch`. I wonder if doing `npm run build` outputs something different?

src/mixins/map-mixin.js
clhenrick: I _think_ this function will only be called when the basemap has been loaded, so not sure that will return `null`. Did you find a case where it was?

biovisualize: No, just curious. I'm barely looking at the logic and seeing that it's technically not impossible that the loop doesn't find any layers[i].type of type "symbol". But it doesn't seem relevant as it works and we can be confident that it finds a match anyway

PR #224, opened 2018-05-30
[WIP] MapD3 and Crossfilter Example

clhenrick:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Author referenced issue(s) fixed by this PR: https://github.com/mapd/mapd-immerse/issues/4496
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


clhenrick:
@biovisualize this is working pretty well but a couple things need to be addressed:

- the axis labels are not populating the chart for some reason despite setting them in the config

- if you brush on the combo chart without applying filters to the other DC charts, but then apply a filter to a DC chart, the brush disappears only to reappear when removing the DC filter. This only seems to happen the very first time and not then after.

example/exampleMapD3Crossfilter.html
```diff
@@ -0,0 +1,745 @@
...
+
+  // date object formatter for constructing SQL query from combo chart brushing
+  const dateFormatter = mapd3.d3.timeFormat("%Y-%m-%d %H:%M:%S")
+
+  // set up a few event listeners for the comboChart's brush events
+  comboChart
+    // we can access the combo chart's event listeners using .getEvents()
+    .getEvents()
+    // note: it's a good idea to "throttle" the brushMove event as it gets called very frequently
+    .onBrush("brushMove", throttle(function() {
```
biovisualize: Good point. We could do it internally as we already do with mousemove.

example/exampleMapD3Crossfilter.html
```diff
@@ -0,0 +1,745 @@
...
+        if (!immediate) func.apply(context, args);
+      };
+      var callNow = immediate && !timeout;
+      clearTimeout(timeout);
+      timeout = setTimeout(later, wait);
+      if (callNow) func.apply(context, args);
+    };
+  };
+
+  function throttle(func, limit) {
```
biovisualize: We could expose the throttle utility function that is in mapd3 src/charts/helpers/common.js.

example/exampleMapD3Crossfilter.html
```diff
@@ -0,0 +1,745 @@
...
+    new MapdCon()
+      .protocol("https")
+      .host("metis.mapd.com")
+      .port("443")
+      .dbName("mapd")
+      .user("mapd")
+      .password("HyperInteractive")
+      .connect(function(error, con) {
+        /*
+         *  This instaniates a new crossfilter.
```
biovisualize: minor typo: instantiates

example/exampleMapD3Crossfilter.html
```diff
@@ -0,0 +1,745 @@
...
+         *  Pass in mapdcon as the first argument to crossfilter, then the
+         *  table name, then a label for the data (unused in this example).
+         *
+         *  to see all availables --  con.getTables()
+         */
+         crossfilter.crossfilter(con, "flights_donotmodify").then(function(cf) {
+           createCharts(cf, con)
+         })
+        /*
+         *  Pass instance of crossfilter into our reateCharts.
```
biovisualize: minor typo: createCharts

example/exampleMapD3Crossfilter.html
clhenrick: true. seems like a sensible default behavior to have built into mapd3.

example/exampleMapD3Crossfilter.html
clhenrick: will fix

example/exampleMapD3Crossfilter.html
clhenrick: will fix

example/exampleMapD3Crossfilter.html
clhenrick: hmmm, we could but it seems that might be going towards the direction of making mapd3 more of a general purpose JS library then a declarative way of creating charts?

PR #225, opened 2018-05-31
Fixes legend locking for chorpleth/pointmap with 'auto' color domains

vastcharade:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/chart-addons/stacked-legend.js
+++ b/src/chart-addons/stacked-legend.js
@@ -45,7 +45,7 @@ function setColorState(setter) {
   }
 }
 
-function setColorScaleDomain(domain) {
+function setColorScaleDomain_v2(domain) {
   return function setState(state) {
     return {
       ...state,
@@ -57,6 +57,21 @@ function setColorScaleDomain(domain) {
             ...state.encoding.color.scale,
             domain
           }
+        }
+      }
+    }
+  }
+}
+
+function setColorScaleDomain_v1(domain) {
+  return function setState(state) {
+    return {
+      ...state,
+      encoding: {
+        ...state.encoding,
+        color: {
+          ...state.encoding.color,
+          domain
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/chart-addons/stacked-legend.js
```diff
@@ -63,6 +63,21 @@ function setColorScaleDomain(domain) {
   }
 }
 
+function setColorScaleDomain_v1(domain) {
```
uyanga-gb: I think we should make the object structure same at some point later for raster charts in immerse and the charting. Needs a minor refactoring.

src/chart-addons/stacked-legend.js
```diff
@@ -135,23 +150,36 @@ export function handleLegendLock({ locked, index = 0 }) {
...
     if (color.legend.locked) {
-      layer.setState(setColorScaleDomain(layer.colorDomain()))
+      layer.setState(setColorScaleDomain_v2(layer.colorDomain()))
     } else {
-      layer.setState(setColorScaleDomain("auto"))
+      layer.setState(setColorScaleDomain_v2("auto"))
+      redraw = true
+    }
+  } else if (typeof color.scale === "undefined" && typeof color.domain !== "undefined") {
+    if (color.legend.locked && color.domain === "auto") {
```
uyanga-gb: There is another property legend.open besides legend.locked, and I am not sure the difference between those two. Does legend.open used somewhere or is it redundant property from immerse? I see legend.open is assigned for Pointmap and Heat map in immerse.

src/chart-addons/stacked-legend.js
vastcharade: absolutely. Not sure why that wasn't done. It's confusing to use two different schemas. Apparently it diverged with heatmap but wasn't brought in line with the other raster layer types. This requires a refactor. See my last comment in this ticket: https://github.com/mapd/mapd-immerse/issues/3703

src/chart-addons/stacked-legend.js
vastcharade: I do not know. Not familiar with the legend code. It seems like it's redundant.

PR #226, opened 2018-06-05
Choropleth API improvements

alexbaden:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -22085,7 +22085,6 @@ function format(_value, _key, numberFormatter, dateFormatter) {
 }
 
 function multipleKeysLabelMixin(_chart) {
-
   function label(d) {
     var numberFormatter = _chart && _chart.valueFormatter();
     var dateFormatter = _chart && _chart.dateFormatter();
@@ -26360,7 +26359,6 @@ function rasterLayerPointMixin(_layer) {
   };
 
   _layer._genVega = function (chart, layerName, group, query) {
-
     // needed to set LastFilteredSize when point map first initialized
     if (_layer.yDim()) {
       _layer.yDim().groupAll().valueAsync().then(function (value) {
@@ -26621,21 +26619,24 @@ function rasterLayerPolyMixin(_layer) {
         layerFilter = _ref$layerFilter === undefined ? [] : _ref$layerFilter,
         filtersInverse = _ref.filtersInverse;
 
-    var selfJoin = state.data[0].table === state.data[1].table;
+    var doJoin = state.data.length > 1;
+    var hasColorMeasure = !(state.encoding.color.domain === undefined);
 
-    var groupby = {
+    var groupby = hasColorMeasure && !(state.data[0].attr === "rowid") ? {
       type: "project",
       expr: state.data[0].table + "." + state.data[0].attr,
       as: "key0"
-    };
+    } : {};
+
+    var rowIdTable = doJoin ? state.data[1].table : state.data[0].table;
 
     var transforms = [{
       type: "rowid",
-      table: state.data[1].table
-    }, !selfJoin && {
+      table: rowIdTable
+    }, doJoin && {
       type: "filter",
       expr: state.data[0].table + "." + state.data[0].attr + " = " + state.data[1].table + "." + state.data[1].attr
-    }, {
+    }, hasColorMeasure && {
       type: "aggregate",
       fields: [layerFilter.length ? _utils.parser.parseExpression({
         type: "case",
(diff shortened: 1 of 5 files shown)


src/mixins/raster-layer-poly-mixin.js
```diff
@@ -180,7 +186,9 @@ export default function rasterLayerPolyMixin(_layer) {
   }
 
   function usesAutoColors() {
-    return state.encoding.color.domain === "auto"
+    return state.encoding.color.domain === undefined
```
jrajav: I'm confused why this change is necessary? state.encoding.color.domain === "auto" will be false if state.encoding.color.domain is undefined, anyway.

dist/mapdc.js
```diff
@@ -26631,21 +26629,24 @@ function rasterLayerPolyMixin(_layer) {
         layerFilter = _ref$layerFilter === undefined ? [] : _ref$layerFilter,
         filtersInverse = _ref.filtersInverse;
 
-    var selfJoin = state.data[0].table === state.data[1].table;
+    var doJoin = state.data.length > 1;
+    var hasColorMeasure = !(state.encoding.color.domain === undefined);
 
-    var groupby = {
+    var groupby = hasColorMeasure && !(state.data[0].attr === "rowid") ? {
```
biovisualize: Maybe `!(state.data[0].attr === "rowid")` could go in a variable to give some info about what it means

dist/mapdc.js
```diff
@@ -26782,15 +26780,31 @@ function rasterLayerPolyMixin(_layer) {
...
-      domain: autocolors ? { data: getStatsLayerName(), fields: ["mincolor", "maxcolor"] } : state.encoding.color.domain,
-      range: colorRange,
-      nullValue: "rgba(214, 215, 214, 0.65)",
-      default: "rgba(214, 215, 214, 0.65)"
-    }];
+    var scales = [];
+    var fillColor = {};
+    var useColorScale = state.encoding.color.value === undefined;
+    if (useColorScale) {
+      var colorRange = state.encoding.color.range.map(function (c) {
```
biovisualize: I guess we are confident that `state.encoding.color.range` will always be populated and will be an array.

dist/mapdc.js
```diff
@@ -26782,15 +26780,31 @@ function rasterLayerPolyMixin(_layer) {
...
+        return (0, _utilsVega.adjustOpacity)(c, state.encoding.color.opacity);
```
biovisualize: I'm not sure I understand how this line works. I guess it's equivalent to `_utilsVega.adjustOpacity(c, state.encoding.color.opacity)`, but what is the purpose of the first `0`?

dist/mapdc.js
```diff
@@ -26839,7 +26850,7 @@ function rasterLayerPolyMixin(_layer) {
   _layer._genVega = function (chart, layerName, group, query) {
     _vega = _layer.__genVega({
       layerName: layerName,
-      filter: _layer.crossfilter().getFilterString(_layer.dimension().getDimensionIndex()),
+      filter: _layer.crossfilter().getFilterString(_layer.dimension() ? _layer.dimension().getDimensionIndex() : null),
```
biovisualize: Maybe the ternary operator could be extracted in a variable

src/mixins/raster-layer-poly-mixin.js
alexbaden: Good catch. Likely a holdover from when `state.encoding.color` could be undefined. I removed it.

src/mixins/raster-layer-poly-mixin.js
```diff
@@ -253,20 +258,36 @@ export default function rasterLayerPolyMixin(_layer) {
...
     }
 
-    const colorScaleName = getColorScaleName(layerName)
-    const scales = [
-      {
+    let scales = []
+    let fillColor = {}
+    const useColorScale = state.encoding.color.value === undefined
+    if (useColorScale) {
+      const colorRange = state.encoding.color.range.map(c =>
```
biovisualize: We could clean-up to be consistent at some point between using `function() {}` or the fat arrow for anonymous functions

dist/mapdc.js
biovisualize: Ah! @jrajav made me realize that I was reviewing the compiled file, which adds these things. So @alexbaden feel free to ignore everything that doesn't relate to an actual code change :)

tmostak:
Hi @alexbaden should we remove mapd_geo from the dimension list now?

PR #231, opened 2018-06-12
filter format when binning is off

biovisualize:
The query was wrong when binning was off because of a bad format of some data attached to the DOM in a mapdc listener

Exposed in Immerse here: https://github.com/mapd/mapd-immerse/pull/4710

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.css
+++ b/dist/mapdc.css
@@ -2198,6 +2198,7 @@ body {
   border: 1px solid white; }
 
 .gradient-legend .block .text {
+  white-space: nowrap;
   position: absolute;
   text-align: center;
   transform: translateX(-50%);
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -52205,7 +52205,7 @@ function mapdTable(parent, chartGroup) {
         }
 
         if (_isGroupedData) {
-          _chart.onClick(d.key0);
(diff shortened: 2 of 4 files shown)


biovisualize:
@clhenrick It works fine with multiple columns. But it doesn't seem to highlight the selected row properly. I will have a look. 
screen shot 2018-06-12 at 4 38 07 pm

clhenrick:
@biovisualize here's the result of my QA:

applying a click filter to the table chart with your fix:

unbin-wtih-chrisv-fix

without your fix:

ubin-without-chrisv-fix

I just pushed an a different solution which seems to work for single un-binned column or multiple columns.

clhenrick:
@biovisualize are you referring to the dist build? I'm guessing that previous PRs didn't rebuild the dist directory which is why there's more stuff in there.

biovisualize:
Right, I fell in this trap again :+1:. I will chase QA to test now then we can merge this

PR #246, opened 2018-07-31
locking heat map legend defaulting min/max right after adjusting 

uyanga-gb:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #[FE-5003]
# Related with 
https://github.com/mapd/mapd-immerse/pull/4891

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/chart-addons/stacked-legend.js
+++ b/src/chart-addons/stacked-legend.js
@@ -196,7 +196,7 @@ export function handleLegendInput({ domain, index = 0 }) {
   const layer = this.getLayers()[index]
   const { scale } = layer.getState().encoding.color
 
-  const colorDomain = layer.getState().encoding.color.domain
+  const colorDomain = layer.getState().encoding.color.domain || layer.getState().encoding.color.scale.domain
   if (colorDomain !== "auto" && _.difference(domain, colorDomain).length > 0) {
     // automatically lock color legend when min/max input changes
     layer.setState(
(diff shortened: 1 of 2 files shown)


src/chart-addons/stacked-legend.js
```diff
@@ -153,7 +153,8 @@ export function handleLegendLock({ locked, index = 0 }) {
   let redraw = false
   if (typeof color.scale === "object") { // this if or raster-layer-heatmap-mixin.js
     if (color.legend.locked) {
-      layer.setState(setColorScaleDomain_v2(layer.colorDomain()))
+      const colorDomain = color.scale.domain ? color.scale.domain : layer.colorDomain()
```
clhenrick: the value of `layer.colorDomain()` could be `null` here, does that matter?

clhenrick: also is passing `color.scale.domain` okay here? It looks like it's passing an array by reference, might be better to clone the array before passing? maybe it doesn't matter.

uyanga-gb: either color.scale.domain or layer.colorDomain() expects an value. In initial null cases, domain should be populated with auto scale min/max.

uyanga-gb: In this case even we pass the array by reference, we not directly modifying the array.

PR #249, opened 2018-08-25
[WIP] Linemap raster chart 

uyanga-gb:
## Description
Adding a new raster chart, Linemap, in immerse that support rendering linestring from mapd_geo column. The below issue is the main issue, and there are several related issues linked to that.
To test the chart, use the the build http://kali.immerse-4915.dev.frontend.builds.mapd.com and use the dataset named `kaggle_taxi_waypoints` or `uyanga_fault_reimported`

## :wrench: Issue(s) fixed:
- [ ] Fixes [FE-4883](https://mapdgpu.atlassian.net/browse/FE-4883)

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


src/mixins/map-mixin.js
```diff
@@ -196,6 +198,17 @@ export default function mapMixin(
     return _chart
   }
 
+  _chart.polyDim = function(polyDim) {
```
jrajav: Can polyDim be a function here? Will it cause any problems if it is?

src/mixins/raster-layer-line-mixin.js
```diff
@@ -0,0 +1,869 @@
...
+    }
+
+    project.setClamp = clamp => {
+      _clamp = Boolean(clamp)
+    }
+
+    return project
+  }
+
+  class GeoSvgFormatter extends PolySvgFormatter {
```
vastcharade: Looks like this is exactly the same as in the poly layer. Can we move this into a utility module somewhere to reduce duplicate code?

src/mixins/raster-layer-line-mixin.js
```diff
@@ -0,0 +1,869 @@
...
+    data,
+    width,
+    height,
+    margins,
+    xscale,
+    yscale,
+    minPopupArea,
+    animate
+  ) {
+    let geoPathFormatter = null
```
vastcharade: Looks like this function is mostly similar to the poly version too except for some subtle differences with the SVG building. Perhaps we can coalesce some of this code into a utility function?

src/mixins/map-mixin.js
```diff
@@ -329,6 +342,16 @@ export default function mapMixin(
             ydim.filter([_chart._minCoord[1], _chart._maxCoord[1]])
           }
         }
+        else if(typeof layer.polyDim === "function") {
+          const polyDim = layer.polyDim()
+          if(polyDim !== null) {
+            redrawall = true
+            polyDim.filterST_Intersects([[bounds._sw.lng, bounds._sw.lat],
```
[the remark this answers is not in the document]
uyanga-gb: @vastcharade Yeah, I agree. I have performance concern on linemap when using either ST_Contains or ST_Intersects as well. What would be a less expensive query to filter on bounds? I haven't actually tried to compare the one you mentioned in Slack because of the caveat to have false positives.

src/mixins/raster-layer-line-mixin.js
```diff
@@ -0,0 +1,869 @@
...
+// NOTE: Reqd until ST_Transform supported on projection columns
+function conv4326To900913(x, y) {
+  const transCoord = [0.0, 0.0]
+  transCoord[0] = x * 111319.49077777777778
+  transCoord[1] =
+    Math.log(Math.tan((90.0 + y) * 0.00872664625997)) * 6378136.99911215736947
+  return transCoord
+}
+
+const AUTOSIZE_DOMAIN_DEFAULTS = [100000, 0]
```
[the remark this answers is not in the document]
uyanga-gb: Yes the max limit is 10k. Should we increase it?

src/mixins/map-mixin.js
uyanga-gb: @jrajav polyDim is a function.

src/mixins/raster-layer-line-mixin.js
```diff
@@ -0,0 +1,869 @@
...
+    }
+
+    return {
+      data,
+      scales,
+      marks
+    }
+  }
+
+  _layer.polyDim = createRasterLayerGetterSetter(_layer, null)
```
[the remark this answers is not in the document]
uyanga-gb: yes it is. Changed

src/mixins/raster-layer-line-mixin.js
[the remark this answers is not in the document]
vastcharade: I'd say bump it up to 100K for now, tho we'll likely want to increase it to several million soon after we land some performance optimizations.

src/utils/utils-vega.js
```diff
@@ -193,3 +194,448 @@ export function createRasterLayerGetterSetter(
...
+// NOTE: Reqd until ST_Transform supported on projection columns
+function conv4326To900913(x, y) {
+  const transCoord = [0.0, 0.0]
+  transCoord[0] = x * 111319.49077777777778
+  transCoord[1] =
+    Math.log(Math.tan((90.0 + y) * 0.00872664625997)) * 6378136.99911215736947
+  return transCoord
+}
+
+class PolySvgFormatter {
```
vastcharade: I'd suggest renaming this base class SvgFormatter

src/utils/utils-vega.js
```diff
@@ -193,3 +194,448 @@ export function createRasterLayerGetterSetter(
...
+   * @param {string} out Returns the svg path string
+   * @param {Point2d} t Translation from world to object space.
+   * @param {Number} s Scale from world to object space.
+   */
+  getSvgPath(t, s) {
+    throw new Error("This must be overridden")
+  }
+}
+
+class LegacySvgFormatter extends PolySvgFormatter {
```
vastcharade: This legacy formatter is for polys only, so let's rename this class LegacyPolySvgFormatter.
It's for backwards compatibility, but could be removed if > 4.x releases of immerse cannot hit a 3.x server.

src/utils/utils-vega.js
```diff
@@ -193,3 +194,448 @@ export function createRasterLayerGetterSetter(
...
+  let geoPathFormatter = null
+  if (chart._useGeoTypes) {
+    if (!state.encoding.geocol) {
+      throw new Error(
+        "No poly/multipolygon column specified. Cannot build poly outline popup."
+      )
+    }
+    geoPathFormatter = new GeoSvgFormatter(state.encoding.geocol)
+  } else {
+    geoPathFormatter = new LegacySvgFormatter()
```
vastcharade: this legacy path is poly only, so let's check for polys (if possible), or throw an error otherwise.
Or, see my previous comment about removing this if different immerse/core versions isn't possible.

src/utils/utils-vega.js
```diff
@@ -193,3 +194,448 @@ export function createRasterLayerGetterSetter(
...
+  // now apply the styles
+  if (typeof strokeWidth === "number") {
+    group.style("stroke-width", strokeWidth)
+  }
+
+  // applying shadow
+  const defs = group
+    .append('defs')
+
+  const filter = defs.append("filter")
```
vastcharade: does this mean that poly pops now have drop shadows too? I'm wondering if it might be good to apply a style config to this function to turn on/off drop shadows, animation, etc. Or just handle this styling in CSS.

uyanga-gb: yes it was, but I removed it from poly.

src/mixins/raster-layer-line-mixin.js
```diff
@@ -0,0 +1,869 @@
...
+            },
+            y: {
+              field: "y"
+            },
+            strokeColor: getColor(state.encoding.color, layerName),
+            opacity: state.mark.opacity, // gets updated when opacity slider changes
+            strokeWidth: size,
+            lineJoin:
+              typeof state.mark === "object" ? state.mark.lineJoin : "miter",
+            miterLimit:
```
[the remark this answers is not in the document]
uyanga-gb: @vastcharade, `lineJoin` is defined as `miter` in immerse. ` so miterLimit is always there. 
mark: {
      type: "lines",
      opacity:
        typeof opacity === "undefined" ? layerDefaultOpacity(type) : opacity,
      lineJoin: "miter",
      miterLimit: 10
    }`

src/mixins/raster-layer-line-mixin.js
```diff
@@ -0,0 +1,869 @@
...
+          type: "formula",
+          expr: "min(maxcol, avgcol+2*stdcol)",
+          as: maxoutput
+        }
+      )
+      rtnobj.fields = [minoutput, maxoutput]
+    } else if (state.encoding.color.type === "ordinal") {
+      const output = "distinctcolor"
+      aggregateNode.fields.push("color")
+      aggregateNode.ops.push("distinct")
```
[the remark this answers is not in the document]
uyanga-gb: We can revisit when we support `auto` for `ordinal` color type. Right now, this block of code will not be used because we only support `auto` for  `quantitative` type. This part is basically copied from point.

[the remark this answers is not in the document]
vastcharade: oh, right. which means I probably added it to points as a placeholder -- just an FYI then.

src/mixins/raster-layer-line-mixin.js
```diff
@@ -0,0 +1,869 @@
...
+    )
+      .filter(
+        transform =>
+          transform.type === "project" && transform.hasOwnProperty("as")
+      )
+      .map(projection => parser.parseTransform({ select: [] }, projection))
+      .map(sql => sql.select[0])
+  }
+
+  function getAutoColorVegaTransforms(aggregateNode) {
```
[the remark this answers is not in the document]
uyanga-gb: @vastcharade, it is modified now.

src/charts/raster-chart.js
```diff
@@ -353,10 +353,12 @@ export default function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
       const layers = getLayers()
       const xRanges = []
       const yRanges = []
+      const polyRanges = []
```
vastcharade: rename `viewRanges` or `viewBoxRanges` or similar

vastcharade: Actually, it doesn't look like polyRanges is used anywhere. Is this a placeholder or can it be removed?

uyanga-gb: @vastcharade it is removed now.

vastcharade:
You're welcome. Thanks for the good work on this anticipated feature!

src/mixins/raster-layer-line-mixin.js
```diff
@@ -212,6 +217,28 @@ function getScales({ size, color }, layerName, scaleDomainFields, xformDataSourc
     })
   }
 
+  if (typeof color === "object" && color.type === "density") {
```
vastcharade: I haven't looked at the code that generates the accumulation scale for points/symbols, but I'm wondering if that code can be moved into a utility function. I'm thinking this code is pretty much the same.

uyanga-gb: @vastcharade, I moved the getScales function to `utils-vega.js`. The logic in the function is almost the same for pointmap and linemap except one minor caveat that pointmap opacity slider controls rgb color opacity for points whereas linemap opacity is slider changes the opacity for the vega layer. We will probably change the opacity control for pointmap later.

vastcharade: ok

PR #260, opened 2018-10-24
[FE-7314], [FE-7313], LINEMAP: run groupby on dimension 

uyanga-gb:
# Merge Checklist
## Description
Support groupby on dimension only without geo join.
## :wrench: Issue(s) fixed:
- [ ] Fixes https://jira.omnisci.com/browse/FE-7314 AND https://jira.omnisci.com/browse/FE-7313

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/raster-layer-line-mixin.js
+++ b/src/mixins/raster-layer-line-mixin.js
@@ -98,7 +98,7 @@ function getTransforms(
     type: "project",
     expr: `${state.data[0].table}.${g}`,
     as: `key${i}`
-  })) : {}
+  })) : []
 
   const groupby = doJoin()
     ? [{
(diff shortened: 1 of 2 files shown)


src/mixins/raster-layer-line-mixin.js
```diff
@@ -94,8 +94,22 @@ function getTransforms(
     return state.data.length > 1
   }
 
+  const groupbyDim = state.transform.groupby ? state.transform.groupby.map((g, i) => ({
+    type: "project",
+    expr: `${state.data[0].table}.${g}`,
+    as: `key${i}`
+  })) : {}
```
jrajav: It seems like groupbyDim (and thus groupby below) can be an object in some cases, yet we're checking the `.length` of it below. This is the only part that seems a little odd. Should this be `[]` instead, perhaps?

uyanga-gb: @jrajav, good catch. I changed. Thank you :)

PR #269, opened 2018-11-09
fix null dim handling and other label issues

biovisualize:
https://jira.omnisci.com/browse/FE-7325

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -47099,6 +47099,11 @@ function pieChart(parent, chartGroup) {
     var arc = buildArcs();
     var centroid = labelCentroid(d, arc);
     var adjacent = Math.abs(centroid[1]);
+
+    if (angle >= Math.TWO_PI) {
+      // when there's a single slice, no need to compute more than that
+      return adjacent;
+    }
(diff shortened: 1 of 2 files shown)


dist/mapdc.js
```diff
@@ -64992,7 +64989,7 @@ function renderNominalLegend(state, dispatch) {
                     h_1.default("div.color", {
                         style: { background: state.range[index] }
                     }),
-                    h_1.default("div.text", "" + value)
+                    h_1.default("div.text", value)
```
int3h: [Non-blocking comment] Out of curiosity, why the removal of `"" + value`? (Not saying that it's wrong, but just want to make sure that wasn't accidental. I believe that exists as an idiomatic way to ensure `value` is casted to a string, which may or may not be necessary if we know for sure `value` is always a string and/or `h_1.default()` doesn't mind if its arguments aren't strings.)

biovisualize: I think this was unrelated code in the map legend, it was just not properly compiled. But `"" + whatever` is a bad way to cast something to a string. It's more readable and robust to do `String(value)` or even `value.toString()`, which would get done anyway since it seem to be assigning some text to a div text. Anyway.

PR #276, opened 2018-11-27
[FE-7268] delete legend when delete layer and handle master layer legend update

uyanga-gb:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes https://jira.omnisci.com/browse/FE-7268

## Relates with 
https://github.com/omnisci/legendables/pull/6

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -64924,7 +64924,6 @@ function bindEventHandlers(chart, container, dataBounds, dataScale, dataOffset, 
 
 "use strict";
 
-
 Object.defineProperty(exports, "__esModule", { value: true });
 var legend_1 = __webpack_require__(207);
 exports.Legend = legend_1.default;
@@ -64936,11 +64935,11 @@ exports.Legend = legend_1.default;
 
 "use strict";
 
-
-var __assign = this && this.__assign || Object.assign || function (t) {
+var __assign = (this && this.__assign) || Object.assign || function(t) {
     for (var s, i = 1, n = arguments.length; i < n; i++) {
         s = arguments[i];
-        for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p)) t[p] = s[p];
+        for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p))
+            t[p] = s[p];
     }
     return t;
 };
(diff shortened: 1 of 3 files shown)


dist/mapdc.js
```diff
@@ -48641,6 +48646,31 @@ function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
     return _legend;
   };
 
+  _chart.deleteLayerLegend = function (currentLayer, deleteLayer, prevLayerId) {
```
biovisualize: I'm guessing that `currentLayer` and `deleteLayer` are IDs like `prevLayerId`. If thats the case, maybe they should be named `currentLayerID` and such.

dist/mapdc.js
```diff
@@ -48641,6 +48646,31 @@ function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
...
+    var legend = document.getElementsByClassName('legend');
+    if (legend.length && currentLayer !== "master") {
+      // the case that has 3 or more layers will be handled in getLegendStateFromChart function in stacked-legend.js by removing extra legend
+      if (prevLayerId && prevLayerId === "master") {
```
biovisualize: Could this be simplified to prevent using an empty return and to merge the 2 `legend[deleteLayer].remove();` cases? Maybe something like:

```
if (prevLayerId && prevLayerId === "master" ||
  (!(currentLayer !== deleteLayer && prevLayerId < deleteLayer) && legend[deleteLayer])) {
  legend[deleteLayer].remove();
} else if (deleteLayer && deleteLayer !== 'master' && legend.length === 1 && legend[0]) {
  legend[0].remove();
}
```

dist/mapdc.js
```diff
@@ -48641,6 +48646,31 @@ function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
...
+      } else if (legend[deleteLayer]) {
+        legend[deleteLayer].remove();
+      } else if (deleteLayer && deleteLayer !== 'master' && legend.length === 1 && legend[0]) {
+        legend[0].remove();
+      }
+    }
+  };
+
+  _chart.destroyChartLegend = function () {
+    var legend = document.getElementsByClassName('legend');
```
biovisualize: Is it possible that this could be simplified as:
```
_chart.selectAll('legend').remove()
```
D3 will remove them only if they exist, and will remove them all.

dist/mapdc.js
```diff
@@ -48641,6 +48646,31 @@ function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
     return _legend;
   };
 
+  _chart.deleteLayerLegend = function (currentLayer, deleteLayer, prevLayerId) {
+    var legend = document.getElementsByClassName('legend');
```
biovisualize: Should we be worried that starting the DOW tree search for `.legend` from the `document` could grab any other `.legend` element in the DOM outside of this chart? I think it should be scoped, something like `_chart.select(".legend")`

dist/mapdc.js
```diff
@@ -48807,20 +48837,36 @@ function setColorScaleDomain_v1(domain) {
...
   return toLegendState(chart.getLayerNames().map(function (layerName) {
     var layer = chart.getLayer(layerName);
+    var layerState = layer.getState();
     var color = layer.getState().encoding.color;
-    if (_typeof(color.scale) === "object" && color.scale.domain === "auto") {
-      return _extends({}, color, {
-        scale: _extends({}, color.scale, {
+
+    if (layers.length > 1 || _.isEqual(selectedLayer, layerState)) {
+      if (_typeof(color.scale) === "object" && color.scale.domain === "auto") {
```
biovisualize: Just a note: I know that's just how the code was before, but I'm wondering what `_typeof()` is doing that `typeof` can't do. And `typeof` is used just a few line under.

dist/mapdc.js
uyanga-gb: @biovisualize, thanks for considering that. It probably does that. I will fix it.

dist/mapdc.js
uyanga-gb: simplified the statement almost as same as you suggested @biovisualize  ;)

src/chart-addons/stacked-legend.js
```diff
@@ -82,26 +82,53 @@ function setColorScaleDomain_v1(domain) {
   }
 }
 
-export function getLegendStateFromChart(chart, useMap) {
+export function getLegendStateFromChart(chart, useMap, selectedLayer) {
+
```
uyanga-gb: @biovisualize,  I was trying to get only relevant chart state by calling `const state = getLegendStateFromChart(_chart, useMap, selectedLayer)` only for current selected layer for the chart, but `_chart._doRender` is called from many different options. Thus, it is difficult to capture currentLayer from immerse and mapd-charting in all those cases. For this reason, only option to remove extra legends is removing in this funtion.

biovisualize:
I still see a lot of `document.getElementsByClassName` that may cause problems.

dist/mapdc.js
uyanga-gb: @biovisualize, that doesn't remove the element from the DOM. I tried and didn't work out.

uyanga-gb:
@biovisualize, I don't see any other issue except the one you posted in the immerse PR. The issue is with **x** button that is not calling the fixed function. Let me know if you find any other bug/bugs.

src/chart-addons/stacked-legend.js
```diff
@@ -82,26 +82,53 @@ function setColorScaleDomain_v1(domain) {
...
+  // the getLegendStateFromChart in _doRender gets called from few different options
+  // and some of them are calling with all layers in chart.
+  // As a result, a legend for each layer is rendered.
+  // Thus, we need to remove extra legends here
+  const allDOMLLegend = document.getElementsByClassName('legend')
+  const layers = chart.getLayerNames()
+  const currentChartId = chart.selectAll('.legend')[0].parentNode.id
+
+  const chartLegends = _.filter(allDOMLLegend, (l) =>
```
biovisualize: You would get the same thing in a more robust way by doing 
```
const chartLegends = chart.root().node().getElementsByClassName('legend')
```

PR #284, opened 2018-12-18
[FE-4902] Update for mapd-connector changes

jrajav:
Update to use new renderVegaAsync method and updated interface to getResultRowForPixel from https://github.com/omnisci/mapd-connector/pull/122


uyanga-gb:
@jrajav, I believe we can test the Examples after you merge it to master then update gh-pages.

PR #291, opened 2018-12-28
[FE-7720] Heat chart - swap selector registers extra chart

uyanga-gb:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes https://jira.omnisci.com/browse/FE-7720

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/base-mixin.js
+++ b/src/mixins/base-mixin.js
@@ -484,6 +484,7 @@ export default function baseMixin(_chart) {
       }
       _root = d3.select(_anchor)
       _root.classed(constants.CHART_CLASS, true)
+      deregisterChart(_chart, chartGroup)
       registerChart(_chart, chartGroup)
       _isChild = false
     } else {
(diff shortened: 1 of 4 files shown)


src/mixins/base-mixin.js
```diff
@@ -484,6 +484,7 @@ export default function baseMixin(_chart) {
       }
       _root = d3.select(_anchor)
       _root.classed(constants.CHART_CLASS, true)
+      deregisterChart(_chart, chartGroup)
```
jrajav: This is in the base-mixin - does this bug actually apply to every chart?

uyanga-gb: @jrajav, as I noticed, this bug also exists in Pie chart. I checked couple other charts which we can switch dimensions such as Bar chart but the bug is not there. I wonder if it is protected somewhere. Anyway, the above function only deregister the current chart, then we register it right after, so it should be safe. I will ask QA to test other thoroughly.

PR #302, opened 2019-02-08
[FE-7750] Fix inconsistent display of mapbox logo / watermark 

CaitW:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes [FE-7750](https://jira.omnisci.com/browse/FE-7750)

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/map-mixin.js
+++ b/src/mixins/map-mixin.js
@@ -278,7 +278,7 @@ export default function mapMixin(
     mapboxlogo.innerHTML = "Mapbox"
 
     const existingLogo = document.getElementsByClassName('mapbox-maplogo')
-    if(existingLogo.length  ) {
+    if(!existingLogo.length) {
       _chart.root()[0][0].appendChild(mapboxlogo)
     }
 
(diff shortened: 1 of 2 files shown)


src/mixins/map-mixin.js
```diff
@@ -277,8 +277,9 @@ export default function mapMixin(
     mapboxlogo.target = "_blank"
     mapboxlogo.innerHTML = "Mapbox"
 
-    const existingLogo = document.getElementsByClassName('mapbox-maplogo')
-    if(existingLogo.length  ) {
+    const thisMapContainer = _map._container
```
jrajav: How certain are we `_map` will never still be null here?

src/mixins/map-mixin.js
```diff
@@ -277,8 +277,9 @@ export default function mapMixin(
...
+    const existingLogo = thisMapContainer.querySelectorAll('.mapbox-maplogo')
```
jrajav: Seems like we could easily just use `querySelector` instead here, but unlikely to matter

CaitW: Changed because `querySelector` is better style

src/mixins/map-mixin.js
CaitW: Added check for null `_map`

PR #313, opened 2019-03-29
[FE-8084] Disable map rotation

israelvicars:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Author referenced issue(s) fixed by this PR: [FE-8084](https://jira.omnisci.com/browse/FE-8084)

## :smoking: Smoke Test
- [x] Works in chrome
- [x] Works in firefox
- [x] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.css
+++ b/dist/mapdc.css
@@ -867,7 +867,7 @@ body {
     box-shadow: 0 1px 4px 0 rgba(0, 0, 0, 0.4);
     pointer-events: none; }
   .dc-chart .map-polygon-shape {
-    pointer-events: auto; }
+    pointer-events: none; }
   .dc-chart .map-popup, .dc-chart .map-point-wrap {
     width: 100%;
     height: 100%;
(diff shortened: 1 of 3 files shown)


dist/mapdc.css
```diff
@@ -867,7 +867,7 @@ body {
     box-shadow: 0 1px 4px 0 rgba(0, 0, 0, 0.4);
     pointer-events: none; }
   .dc-chart .map-polygon-shape {
-    pointer-events: auto; }
+    pointer-events: none; }
```
biovisualize: Can you just document what this solves?

israelvicars:
Corresponding Immerse PR for testing: https://github.com/omnisci/mapd-immerse/pull/5324

dist/mapdc.css
israelvicars: Noting as a comment here that it disables map rotation by preventing the pointer-events for the object with this class. When map rotation is not disabled, right-click dragging is the relevant event source on desktop and pinch on touch-enabled devices.

PR #314, opened 2019-03-31
[FE-8203]  filter on aggregate pointmap

uyanga-gb:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes https://jira.omnisci.com/browse/FE-8203

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


src/mixins/raster-layer-point-mixin.js
```diff
@@ -199,6 +225,20 @@ function getTransforms(
     })
   }
 
+  const postFilter = postFilters ? postFilters[0] : null // may change to map when we have more than one postFilter
+  if (postFilter && isValidPostFilter(postFilter)) {
```
jrajav: So if we have an invalid postFilter, we just silently ignore it? Is that ever expected to happen, in normal operation?

uyanga-gb: it is basically checks relation between operator and min or max values and the the value type. If it is not satisfies the minimal requirement, yes we ignore the filter get applied.

src/mixins/raster-layer-point-mixin.js
```diff
@@ -96,11 +96,37 @@ function getColor(color, layerName) {
...
+    table,
+    custom
+  } = postFilter
+
+  if (value && (aggType || custom)) {
+    if ((operator === "not between" || operator === "between") && (typeof min === "number" && !isNaN(min)) && (typeof max === "number" && !isNaN(max))) {
+      return true
+    } else if ((operator === "equals" || operator === "greater than") && (typeof min === "number" && !isNaN(min))) {
+      return true
+    } else if (operator === "greater than" && typeof min === "number" && !isNaN(min)) {
```
bmatcuk: Isn't this case covered by the "else if" above?

uyanga-gb: @bmatcuk, yes it is. Good catch. Thank you.

PR #320, opened 2019-04-10
[FE-8329][WIP] Disable render/redraw stack guards

int3h:
In `core-async.js` in mapdc, in `redrawAllAsync()` and `renderAllAsync()`, there are checks for `if (!stackEmpty) {...}` that throw away render/redraw calls in-flight when another render/redraw call is requested. This causes a number of bugs, including FE-7800, FE-7815, and FE-8326, which all revolve around renders not happening correctly.

We should try removing this logic and verify that no bugs are introduced. Specifically, we should verify that the original "infinite render loops" bugs that caused this logic to be added are not still present when this logic is removed.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes FE-8329

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -4773,14 +4773,17 @@ function refreshDisabled(_) {
   if (!arguments.length) {
     return _refreshDisabled;
   }
+  console.debug(">>> Setting refresh disabled to:", _);
   _refreshDisabled = _;
 }
 
 function disableRefresh() {
+  console.debug(">>> Disabling refresh");
   _refreshDisabled = true;
 }
 
 function enableRefresh() {
+  console.debug(">>> Enabling refresh");
   _refreshDisabled = false;
 }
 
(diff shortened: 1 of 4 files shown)


mrblueblue:
For context, this logic was put in place specifically for the needs of the line chart, which crossfilters and redraws its entire chart group through "brushing". 

The whole redrawAll/renderAll might be over-engineered and was basically hacked to accommodate the raster charts which need to query for the total number of records before making it's main query. 

I think the whole logic is roughly: do not render/redraw if the group is currently redrawing/ rendering and if a render/redraw attempt happens during a pending cycle, then call render/redraw again after the cycle is finished.

PR #322, opened 2019-04-11
[FE-7408] fixed undefined return value

bmatcuk:
If the domain of a quant scale is `[x, x]`, the return value of `scale(x)` will be undefined since both the low-end and high-end range values are equally as valid. Simple fix to return the default value if the result is undefined.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes https://jira.omnisci.com/browse/FE-7408

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/utils/utils-vega.js
+++ b/src/utils/utils-vega.js
@@ -160,6 +160,10 @@ export function createVegaAttrMixin(
         } else {
           quantScale.domain(domainVals).range(capAttrObj.range)
           rtnVal = quantScale(input)
+
+          if (typeof rtnVal === "undefined") {
+            rtnVal = layerObj[defaultFunc]()
+          }
         }
       }
     }


src/utils/utils-vega.js
```diff
@@ -160,6 +160,10 @@ export function createVegaAttrMixin(
         } else {
           quantScale.domain(domainVals).range(capAttrObj.range)
           rtnVal = quantScale(input)
+
+          if (typeof rtnVal === "undefined") {
+            rtnVal = layerObj[defaultFunc]()
```
uyanga-gb: @bmatcuk, I don't think we can use default size here. This bug is caused by setting the SIZE DOMAIN min/max value to the same value which caused d3 quantize scaling got infinity value. There is SIZE RANGE slider just underneath the SIZE DOMAIN which is tightly dependent one another. Assuming user changed the SIZE RANGE to some other value other than default min 3, then setting the SIZE DOMAIN to min=1 max=1 as in the issue, then your fix with default would be mismatch. I think we can use different scale, but @biovisualize  is our d3 expert.

bmatcuk:
Oh, I was assuming the "default" was just the minimum of the range... if that's not true, why don't I just change the code to use the minimum of the range? ie, `range[0]`? That should solve the problem. What do you think?

uyanga-gb:
We can use the min range to fix the bug, but I am not entirely sure if that is a good UX.

biovisualize:
We decided with @bmatcuk to close this one and use this alternative instead https://github.com/omnisci/mapd-charting/pull/323

PR #323, opened 2019-04-12
Tooltip offset problem

biovisualize:
Alternative fix for FE-7408. Replaces quantize scale by linear when domain min/max are the same; Also detects the "Other" color item so it uses the proper default "Other" color.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #FE-7408

Also fixes an unfiled bug where color for Other categories are not blue:
Before:
Capture d’écran, le 2019-04-12 à 15 29 59

After:
Capture d’écran, le 2019-04-12 à 16 08 18

### DIFF
--- a/src/utils/utils-vega.js
+++ b/src/utils/utils-vega.js
@@ -32,6 +32,7 @@ export function adjustRGBAOpacity(rgba, opacity) {
 
 const ordScale = d3.scale.ordinal()
 const quantScale = d3.scale.quantize()
+const linearScale = d3.scale.linear()
 
 const capAttrMap = {
   FillColor: "color",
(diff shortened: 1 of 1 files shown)


src/utils/utils-vega.js
```diff
@@ -108,7 +109,7 @@ export function createVegaAttrMixin(
         const colorScaleName = layerName + "_" + attrName,
           colorsToUse = layerObj.defaultFillColor(),
           domainInterval = 100 / (colorsToUse.length - 1),
-          linearScale = colorsToUse.map((color, i) => i * domainInterval / 100),
+          linearColorScale = colorsToUse.map((color, i) => i * domainInterval / 100),
```
biovisualize: Here I just renamed a "linearScale" to be more specific, to not just let it be shadowed by the new one above

src/utils/utils-vega.js
```diff
@@ -156,7 +157,12 @@ export function createVegaAttrMixin(
         }
         if (capAttrObj.type === "ordinal") {
           ordScale.domain(domainVals).range(capAttrObj.range)
-          rtnVal = ordScale(input)
+          // if color range is not in domain, it's an Other item
+          rtnVal = domainVals.indexOf(input) === -1 ? ordScale("Other") : ordScale(input)
```
biovisualize: It's probably safe to count on the "Other" item to be in the range list anytime you have an input that is not in it as it means that the "Other" feature is toggled on in the color palette widget. Anyway if for any reason it's not there, it will default to the first item from the domain as it was doing before.

src/utils/utils-vega.js
```diff
@@ -156,7 +157,12 @@ export function createVegaAttrMixin(
...
+        } else if (Array.isArray(domainVals) && domainVals[0] === domainVals[1]) {
```
uyanga-gb: Seems reasonable to use linear scale in this case.

PR #325, opened 2019-04-23
Number chart dynamic padding

biovisualize:
https://jira.omnisci.com/browse/FE-4920


dist/mapdc.js
```diff
@@ -68743,10 +68743,12 @@ function numberChart(parent, chartGroup) {
...
-    var TEXT_MARGINS = 64;
-    var chartWidth = _chart.width() - TEXT_MARGINS;
-    var chartHeight = _chart.height() - TEXT_MARGINS;
-    var fontSize = _utils.utils.getFontSizeFromWidth(formattedValue, chartWidth, chartHeight);
+    var TEXT_PADDING_RATIO = 5;
+    var chartWidth = _chart.width();
+    var chartHeight = _chart.height();
+    var wrapperWidth = chartWidth - chartWidth / 100 * TEXT_PADDING_RATIO;
+    var wrapperHeight = chartHeight - chartHeight / 100 * TEXT_PADDING_RATIO;
+    var fontSize = _utils.utils.getFontSizeFromWidth(formattedValue, wrapperWidth, wrapperHeight);
```
RavenHursT: Just wondering if maybe this could somehow be accomplished w/ CSS `calc()`? 🤔

biovisualize: I'm indeed used to just use a responsive SVG. But if I have to do more than a few line changes to this widget, I would prefer to extract it from mapdc and make a better one in React, with features that were requested before like delta changes and conditional coloring.

RavenHursT: Sounds like a GREAT idea ;-)

That's what I did in regards to the legend.  If we know that's what's best. Scope it out and just do it :-)

PR #326, opened 2019-04-26
[FE-7857] block/delay repeated render/redraw requests *by group*

bmatcuk:
Currently, `renderAllAsync` and `redrawAllAsync` throw away any requests to render/redraw if it's currently rendering/redrawing something. This PR changes the logic to keep track of requests per group:
* If the group is *not* rendering/redrawing, then it starts, even if other group(s) are rendering/redrawing;
* Otherwise,
  * If a "note" has been made (see next bullet), nothing happens; request is thrown away.
  * If a "note" has not been made, then it is noted that the group has requested another render/redraw. When the original render/redraw has finished, the note is deleted and the group renders/redraws again.

All of this logic assumes the `allCharts` parameter is falsey. If it is truthy, then that usurps all the logic:
* If nothing is rendering/redrawing currently, everything renders/redraws,
* Otherwise, a "note" is made that everything should render/redraw. As long as this note exists, any additional requests to render/redraw (even by group) are thrown away. When the original render/redraw finishes, the note is deleted and everything renders/redraws.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes [FE-7857](https://jira.omnisci.com/browse/FE-7857)
  - Also relates to [FE-8329](https://jira.omnisci.com/browse/FE-8329)

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/core/core-async.js
+++ b/src/core/core-async.js
@@ -4,8 +4,6 @@ let _renderId = 0
 let _redrawId = 0
 let _renderCount = 0
 let _redrawCount = 0
-let _renderIdStack = null
-let _redrawIdStack = null
 let _renderStackEmpty = true
 let _redrawStackEmpty = true
 let _startRenderTime = null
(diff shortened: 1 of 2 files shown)


src/core/core-async.js
```diff
@@ -14,6 +12,81 @@ let _startRedrawTime = null
...
+  }
+
+  // Utility function to check if a render/redraw should start for the given
+  // group or "all".
+  shouldStart(group, all) {
+    // 1. If currently rendering/redrawing all, return false.
+    // 2. If we're requesting a render/redraw all and *anything* is currently
+    //    rendering/redrawing, return false.
+    // 3. Otherwise, if the requested group is rendering/redrawing, return
+    //    false.
```
RavenHursT: Is there a usecase here where this returns `true`? 🤔

bmatcuk: Oh, every case that's not documented to return false =) I suppose I could add a comment about that... done!

PR #331, opened 2019-04-30
[FE-7857] block/delay repeated render/redraw requests *by group*

bmatcuk:
I reworked a lot of the render/redraw logic in this PR to remove these confusing "stacks". The idea is:

1. If there is a request to render/redraw and nothing is rendering/redrawing, it begins immediately.
2. If "all" charts are already rendering/redrawing, or if the given group is already rendering/redrawing, *and* there is no pending request for all or the group, then the request is marked pending and a promise is returned. Once the original rendering/redrawing has finished, the pending request begins automatically. Once it finishes, the promise will resolve or reject appropriately.
3. If there is already a pending request, the promise for the pending request is returned. In this way, a maximum of one pending request will queue up for the given group (or all), and all callers will be notified when it finishes via the promise.

There's also some logic where a pending "all" request will usurp pending groups; ie, if a pending group is selected to run, but there's a pending "all", it just returns the pending all's promise and exits since the all will take care of this group automatically.

I've also changed the "stack empty" logic: basically, if the given group (or "all") is not rendering or redrawing, the stack is empty. This is simple enough, I think, to prevent the infinite loop this code was originally trying to solve.

This is a big PR, but hopefully it makes more sense than the previous code. I tried to add comments and unit tests to make the code maintainable for future code spelunkers.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes [FE-7857](https://jira.omnisci.com/browse/FE-7857)

## :smoking: Smoke Test
- [x] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -5591,8 +5591,11 @@ utils.nullsLast = function (sorting) {
 Object.defineProperty(exports, "__esModule", {
   value: true
 });
+exports.LockTracker = undefined;
 
 var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
+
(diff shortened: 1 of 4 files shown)


RavenHursT:
in case #1, is a promise returned as well?

bmatcuk:
@RavenHursT yes, a promise is always returned.

RavenHursT:
@bmatcuk ok.. just wondering.. #2 and #3 both mention returning promises, #1, doesn't.

PR #351, opened 2019-06-26
[FE-8750] Anticipate milliseconds when comparing dates in table chart

CaitW:
Dates were being coerced to strings that didn't include milliseconds, so now any dates in filters or test values are coerced to ISO strings when being compared.

Also added documentation and split up some code for readability.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/filter-mixin.js
+++ b/src/mixins/filter-mixin.js
@@ -35,12 +35,7 @@ export function addFilterHandler(filters, filter) {
 export function hasFilterHandler(filters, testValue) {
   if (typeof testValue === "undefined") {
     return filters.length > 0
-  } else {
-    return testIfValueWithinFilters(filters, testValue)
   }
-}
-
-function testIfValueWithinFilters(filters, testValue) {
   testValue = Array.isArray(testValue)
     ? testValue.map(normalizeArrayByValue)
     : testValue
--- a/src/mixins/filter-mixin.js
+++ b/src/mixins/filter-mixin.js
@@ -44,25 +44,28 @@ function testIfValueWithinFilters(filters, testValue) {
   testValue = Array.isArray(testValue)
     ? testValue.map(normalizeArrayByValue)
     : testValue
-  const filtersWithIsoDates = convertAllDatesInArrayToISOString(filters)
-  const testValueWithISODates = convertAllDatesInArrayToISOString(testValue)
+  const filtersWithIsoDates = convertAllDatesToISOString(filters)
+  const testValueWithISODates = convertAllDatesToISOString(testValue)
   return filtersWithIsoDates.some(
     f => testValueWithISODates <= f && testValueWithISODates >= f
   )
 }
 
-// Assuming array can only have one level of nested arrays
-function convertAllDatesInArrayToISOString(a) {
-  return a.map(value => {
-    if (Array.isArray(value)) {
-      return value.map(v => {
-        if (v instanceof Date) {
-          return v.toISOString()
-        }
-      })
-    }
-    return value instanceof Date ? value.toISOString() : value
-  })
+function convertAllDatesToISOString(a) {
+  if (Array.isArray(a)) {
+    // Assuming array can only have up to one level of nested arrays
+    return a.map(value => {
+      if (Array.isArray(value)) {
+        return value.map(v => {
+          if (v instanceof Date) {
+            return v.toISOString()
+          }
+        })
+      }
+      return value instanceof Date ? value.toISOString() : value
+    })
+  }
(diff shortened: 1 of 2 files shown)


src/mixins/filter-mixin.js
```diff
@@ -15,17 +15,56 @@ export function addFilterHandler(filters, filter) {
   return filters
 }
 
-export function hasFilterHandler(filters, filter) {
-  if (typeof filter === "undefined") {
+/**
+ * hasFilterHandler
+ * - if testValue is undefined, checks to see if the chart has any active filters
```
int3h: I *love* these comments. It would take me forever to figure out what was going on without this comment block!

src/mixins/filter-mixin.js
```diff
@@ -15,17 +15,56 @@ export function addFilterHandler(filters, filter) {
...
-  } else if (Array.isArray(filter)) {
-    filter = filter.map(normalizeArrayByValue)
-    return filters.some(f => filter <= f && filter >= f)
   } else {
-    return filters.some(f => filter <= f && filter >= f)
+    return testIfValueWithinFilters(filters, testValue)
   }
 }
 
+function testIfValueWithinFilters(filters, testValue) {
```
int3h: I'm curious why you broke this into a separate function? It leaves the calling function with just a simple `if` statement that dispatches all logic here. Not a blocking comment, just curious about it.

src/mixins/filter-mixin.js
```diff
@@ -15,17 +15,56 @@ export function addFilterHandler(filters, filter) {
...
+/**
+ * hasFilterHandler
+ * - if testValue is undefined, checks to see if the chart has any active filters
+ * - if testValue is defined, checks to see if that testValue passes the active filter
+ *
+ * @param {*} filters - the chart's current filter
+ * @param {*} testValue - a value being tested to see if it passes the filter
+ *
+ *  - If chart values are not binned:
+ *      - Params will be an array of values
```
int3h: Does "params" here mean `testValue`?

src/mixins/filter-mixin.js
```diff
@@ -15,17 +15,56 @@ export function addFilterHandler(filters, filter) {
...
+  const testValueWithISODates = convertAllDatesInArrayToISOString(testValue)
+  return filtersWithIsoDates.some(
+    f => testValueWithISODates <= f && testValueWithISODates >= f
+  )
+}
+
+// Assuming array can only have one level of nested arrays
+function convertAllDatesInArrayToISOString(a) {
+  return a.map(value => {
+    if (Array.isArray(value)) {
```
int3h: May be more elegant to have this function call itself recursively if it's an array of arrays. (Non-blocking, since it works equally well here, but recursion might be more elegant/simpler/more robust).

src/mixins/filter-mixin.js
```diff
@@ -15,17 +15,56 @@ export function addFilterHandler(filters, filter) {
...
   }
 }
 
+function testIfValueWithinFilters(filters, testValue) {
+  testValue = Array.isArray(testValue)
+    ? testValue.map(normalizeArrayByValue)
+    : testValue
+  const filtersWithIsoDates = convertAllDatesInArrayToISOString(filters)
+  const testValueWithISODates = convertAllDatesInArrayToISOString(testValue)
+  return filtersWithIsoDates.some(
```
int3h: I'm probably not understanding things here, but wouldn't you want `all` instead of `some`? That is, aren't you checking if the test value passes every filter, not just one?

CaitW: 🤷‍♂ That's what the original check did? I don't understand it either. I should probably try...

int3h: I know the old code used `some`, so keeping that might be prudent just so not to break any current behavior (even if that behavior is incorrect.... oh Immerse).

src/mixins/filter-mixin.js
CaitW: Both params - filters, and testValue

int3h: So are `filters` and `testValues` always the same structure (either a plain array, or an array of arrays)? Or can one be a regular array while the other is a array of arrays?

CaitW: I only looked into a couple test cases, but that's how it appeared. I added some uncertainty to my documentation just in case

PR #354, opened 2019-07-02
[FE-8727][FE-8787] Fix Heat chart x-axis labels, popups display when specifying custom date format

CaitW:
Custom dimension date formats weren't appearing correctly in heat chart - both in axis labels and popups. 

This is because the functions that created the text for rows / columns in the heat chart were passing along the information in the incorrect format. 

`customFormatter` is set to the `autoFormatter` function from mapd3, which processes raw values. However, `formatDataValue` expects an object / array of objects that contain additional information

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -66905,30 +66905,48 @@ function heatMapValueAccesor(_ref2) {
 
 function heatMapRowsLabel(d) {
   var value = this.rowsMap.get(d) || d;
-  var valueIsFormattableDate = false;
 
+  // There will only return a value if we're displaying dates
   var customFormatter = this.dateFormatter();
+
+  // Possibly dead code, `d` should always be a string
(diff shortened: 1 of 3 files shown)


CaitW:
> Approved with some suggestions:
> 
> * If `d` should always be a string, what about adding a check for that? Or an optional cast if that's required? It would make it safer if that case is being supported in the function.
> * Also,  if you happen to refactor any, it seems like the single letter var `d` would be better as `date` or something more descriptive. Not your code, just mentioning if you decide to come back to it.

As far as I could tell, `d` _should_ always be a string, but I didn't try every possible dimension data type to verify that. I think `d` can also be affected based on binning. Basically I left that code because I was afraid of breaking things by removing it.

`d` is typically the variable name that d3 uses in a lot of its examples, I believe it just stands for "data"

PR #355, opened 2019-07-08
[FE-4958] High Precision Timestamps

CaitW:
[Immerse PR](https://github.com/omnisci/mapd-immerse/pull/5534)


dist/mapdc.js
```diff
@@ -7858,7 +7858,10 @@ var momentUTCFormat = exports.momentUTCFormat = function momentUTCFormat(d, f) {
   return _moment2.default.utc(d).locale("en").format(f);
 };
 var genericDateTimeFormat = exports.genericDateTimeFormat = function genericDateTimeFormat(d) {
-  return momentUTCFormat(d, "MMM D, YYYY") + " \u205F" + momentUTCFormat(d, "HH:mm:ss");
+  if (d.getMilliseconds() === 0) {
```
CaitW: For now, we're truncating millisecond timestamps is the `d.getMilliseconds() === 0`

src/utils/formatting-helpers.js
```diff
@@ -13,8 +13,12 @@ export const momentUTCFormat = (d, f) =>
     .utc(d)
     .locale("en")
     .format(f)
-export const genericDateTimeFormat = d =>
-  `${momentUTCFormat(d, "MMM D, YYYY")} \u205F${momentUTCFormat(d, "HH:mm:ss")}`
+export const genericDateTimeFormat = d => {
+  if (d.getMilliseconds() === 0) {
```
CaitW: For now, we're truncating millisecond timestamps is the `d.getMilliseconds() === 0`

src/charts/line-chart.js
```diff
@@ -449,7 +449,7 @@ export default function lineChart(parent, chartGroup) {
 
       const dots = g
         .selectAll("circle." + DOT_CIRCLE_CLASS)
-        .data(points, pluck("x"))
```
CaitW: pluck does weird things to dates

src/utils/binning-helpers.js
```diff
@@ -62,15 +66,13 @@ export function autoBinParams(timeBounds, maxNumBins, reverse) {
     return DEFAULT_NULL_TIME_RANGE
   }
   const timeSpans = reverse ? TIME_SPANS.slice().reverse() : TIME_SPANS
+
   for (let s = 0; s < timeSpans.length; s++) {
-    if (
-      timeRange / timeSpans[s].numSeconds < maxNumBins &&
-      timeRange / timeSpans[s].numSeconds > 2
-    ) {
+    if (timeRange / timeSpans[s].numSeconds < maxNumBins) {
```
CaitW: See [this PR](https://github.com/omnisci/mapd-charting/pull/352) for this specific change

src/charts/mapd-table.js
```diff
@@ -344,8 +344,14 @@ export default function mapdTable(parent, chartGroup) {
...
-          const key = (val && val[0] && val[0].isExtract) ? null : col.measureName || col.expression
-          return customFormatter && customFormatter(val, key) || formatDataValue(val)
+          const key =
+            val && val[0] && val[0].isExtract
+              ? null
+              : col.measureName || col.expression
+          return (
+            (customFormatter && customFormatter(val, key)) ||
+            formatDataValue(val)
+          )
```
thomasoniii: This is a stylistic suggestion, so don't feel obligated.

What about defining customFormatter as:

`customFormatter ||= _chart.valueFormatter() || (val, key) => formatDataValue(val)`

Then the later logic can be simplified a lot. e.g.:
`(customFormatter && customFormatter(val, key)) || formatDataValue(val)`
becomes:
customFormatter(val, key)

src/mixins/filter-mixin.js
```diff
@@ -15,15 +15,52 @@ export function addFilterHandler(filters, filter) {
...
+          if (v instanceof Date) {
+            return v.toISOString()
+          }
+        })
+      }
+      return value instanceof Date ? value.toISOString() : value
+    })
+  }
+  return a instanceof Date ? a.toISOString() : a
 }
```
thomasoniii: I'm pretty sure there's a logic bug in this function.

convertAllDatesToISOString will return the original values if they're in the top level array.

That is: convertAllDatesToISOString(['a', 'b', 'c']) will return ['a', 'b', 'c'], per the conditional on line 60.

But if you have a nested array of non-date values, it'll return undefined:

That is: convertAllDatesToISOString([['a', 'b', 'c']]) will return [undefined, undefined, undefined], because they won't be caught by the instanceof call on 55, so undefs are returned.

I assume that the conditional on lines 55-57 should just be replaced with the ternary on line 60. Maybe that could be abstracted out further, but meh.

CaitW: There definitely is. Fixed with recursion

thomasoniii: much nicer!

PR #365, opened 2019-07-23
[FE-8836] Fix color legend & bar chart X axis formatting

int3h:
# Summary
### Fix formatting of color dimension

There was a, uh, bug?, where the color legend wasn't passing the color measure name to mapd3's `autoFormatter`. This meant that mapd3 wouldn't be able to associate the color legend values with a specific format string, and would default to the first format string found (for reasons). This commit changes it so that the color legend looks up the color measure name before formatting the legend values.

### Fix formatting of X axis values in row chart

There was a bug where the X axis would only check that _some_ value formatter had been set before trying to format the X axis values, but it didn't check that the formatter had any specific formatter for the X axis. Because mapd3 will default to the first format string found if it doesn't know how to format a specific measure, this caused weirdness when the user set a color format but not a regular measure format.

This commit changes that behavior so that we check the value formatter to ensure it knows how to format the X axis values. Because the value formatter is a block box, we can't test this directly. Instead, we run through a test value (`0.0`) and see if it returns a value value.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes FE-8836

## :smoking: Smoke Test
- [x] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -51551,6 +51551,10 @@ function defaultFormatter(value) {
 
 // Finds the name of the color measure, so we can use it when finding the format of the color legend
 function getColorMeasureName(chart) {
+  var groups = chart.group();
+  if (!(groups && groups.reduce)) {
+    return null;
+  }
   var measures = chart.group().reduce();
   if (!Array.isArray(measures)) {
     return null;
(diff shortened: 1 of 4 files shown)


uyanga-gb:
@int3h, lgtm besides color measure legend with more than one dimension formatting. I am not sure you need to fix that in the issue.

uyanga-gb:
The bug in above comment is fixed https://github.com/omnisci/mapd-immerse/pull/5622

PR #370, opened 2019-07-25
[FE-8934] Fix hpt binning with colors in histogram

int3h:
Use millisecond resolution for Dates in multi-series-mixin

In the `processMultiSeriesResults()` function in `multi-series-mixin.js`, add a check to see if the values be used are Date objects. If so, explicitly call `toISOString()` instead of letting it implicitly call `toString()`, so that we retain millisecond precision.

This change should fix using HPT values on histogram with a color dimension. Previously, you would only get one bin, since `toString()` cuts off milliseconds. Now, we get the proper binning.

This PR also fixes two other functions, `_title()` and `compareDomains()` to be HPT-aware. These weren't directly related to the bug in question, but were incidentally discovered in the course of fixing this bug.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes FE-8934

## :smoking: Smoke Test
- [x] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/base-mixin.js
+++ b/src/mixins/base-mixin.js
@@ -95,7 +95,6 @@ export default function baseMixin(_chart) {
   let _title = function(d) {
     const key = _chart.keyAccessor()(d)
     const value = _chart.valueAccessor()(d)
-    // return _chart.keyAccessor()(d) + ": " + _chart.valueAccessor()(d)
     return `${key instanceof Date ? key.toISOString() : key}: ${
       value instanceof Date ? value.toISOString() : value
     }`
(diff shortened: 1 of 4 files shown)


src/mixins/base-mixin.js
```diff
@@ -93,7 +93,12 @@ export default function baseMixin(_chart) {
   let _renderLabel = false
 
   let _title = function(d) {
-    return _chart.keyAccessor()(d) + ": " + _chart.valueAccessor()(d)
+    const key = _chart.keyAccessor()(d)
+    const value = _chart.valueAccessor()(d)
+    // return _chart.keyAccessor()(d) + ": " + _chart.valueAccessor()(d)
```
israelvicars: @int3h Do you want to remove this comment?

PR #373, opened 2019-07-26
[WIP][FE-8035] Upgrade Mapbox

CaitW:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


vastcharade:
We're going to need to produce un-premultiplied images from the BE, otherwise immerse will be limited to opaque colors only, and there can certainly be scenarios where that is unwanted.

Here's the BE ticket: https://jira.omnisci.com/browse/BE-3956

src/mixins/map-mixin.js
```diff
@@ -2,7 +2,6 @@ import d3 from "d3"
 import * as _ from "lodash"
 import { redrawAllAsync, resetRedrawStack } from "../core/core-async"
 import { utils } from "../utils/utils"
-import { mapDrawMixin } from "./map-draw-mixin"
```
uyanga-gb: Thanks for removing this.

src/mixins/map-mixin.js
```diff
@@ -545,12 +538,17 @@ export default function mapMixin(
         coordinates: boundsToUse
       })
 
+      const mapLayers = _chart.getAllLayers()
+      const chartType =
+        Array.isArray(mapLayers) && mapLayers[0] && mapLayers[0].type
+      const opacity = chartType === "heatmap" ? 0.5 : 0.85
```
uyanga-gb: @CaitW, is that mean we can't go higher than 50% of opacity for the the heatmap and 85% for all the other charts?
We have an opacity adjuster slider in Master layer view and that takes the current opacity value as it is but will allow to increase the opacity to full 100%. Though, with this change, it sets the 50% and 85% to the full 100%, so we can't go up but just lower. Please check the attached image for clarification.
Screen Shot 2019-08-12 at 10 58 21 AM
Screen Shot 2019-08-12 at 11 00 41 AM

CaitW: Yep - you're right. I didn't think about multi-layer maps when I hard-coded the values. I'll add a fix.

uyanga-gb:
If light-v9 is the only option that is comparable to light-v8, we can go with it. Seems like all functionalities are working fine based on my testing. Slight opacity difference in Scatter chart specially, but assuming we are ok with that. Only thing needed to get fixed is the slider value should reflect the current opacity for the multilayers.

uyanga-gb:
One more thing @CaitW, is your immerse build is up to date? Does it include the raster resize fix that REMOVES   `canvas.mapboxgl-canvas {
        width: 100% !important;
 }`
It looks like when we are changing the chart size, the map is still trying to keep the 100% width which causes the skewed map. 
raster_resize

src/charts/raster-chart.js
```diff
@@ -760,6 +760,9 @@ function genLayeredVega(chart) {
   const vegaSpec = {
     width: Math.round(width),
     height: Math.round(height),
+    viewRenderOptions: {
```
uyanga-gb: Is this the trick BE needs?

PR #374, opened 2019-07-29
[FE-8955] Format codebase + add formatting checker to test suite

int3h:
The mapd-charting repo already has a formatter and formatting rules configured for it. However, since the developer must manually run `npm run format`, and this is rarely done, a substantial portion of the files in the repo have incorrect formatting.

The acceptance criteria for this ticket are two-fold:

1. Run `npm run format` on the entire mapd-charting repo to ensure all files are correctly formatted.
2. Add a "check formatting" step to the test suite (i.e. `npm run test` command) so that formatting problems will cause a test failure, and block the PR containing them from being merged.

**Note**: this diff is huge because people haven't run `prettier` in forever, so nearly every file had a change. These should be 100% formatting changes, though, and not change functionality.

Also note to future `git blame` users who have traced back the entire codebase to this PR: I'm sorry, and this isn't the diff you are looking for.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes FE-8955

## :smoking: Smoke Test
- [x] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4919,7 +4919,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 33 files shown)


RavenHursT:
@int3h wondering if it'd be a good idea, as part of this, to create an `immerse-lint-spec` (or something) repo w/ our Immerse `.eslintrc.js` file in it, and then import that into this repo as a `devDependency`, so that the two packages are using the same standard?

int3h:
@RavenHursT that might be nice, though I'm not sure how exactly to do that. Would be happy to have commits to this PR that do that, though!

PR #377, opened 2019-09-09
[FE-9094] expose filterObj

thomasoniii:
This is necessary for [FE-9094](https://jira.omnisci.com/browse/FE-9094).

This change just implments some refactoring for the sole purpose of exposing a `getFilterObj` method on the dcChart. We can use that method to pull out the relevant filter information for display in the pop up.

No functionality should be changing other than the newly exposed method.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -30534,10 +30534,32 @@ function rasterDrawMixin(chart) {
 
     var layers = chart.getLayers && typeof chart.getLayers === "function" ? chart.getLayers() : [chart];
     layers.forEach(function (layer) {
-      if (isLayerTypePointsOrHeatOrUndefined(layer)) {
-        var filterObj = getRasterFilterObj(layer);
+      if (!layer.layerType || typeof layer.layerType !== "function" || layer.layerType() === "points" || layer.layerType() === "heat") {
+        var crossFilter = null;
+        var filterObj = null;
+        var group = layer.group();
 
-        if (filterObj) {
+        if (group) {
+          crossFilter = group.getCrossfilter();
+        } else {
+          var dim = layer.dimension();
+          if (dim) {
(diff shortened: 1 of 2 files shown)


uyanga-gb:
@thomasoniii, as you know we have polygon selection filter for Choropleth chart that is handled https://github.com/omnisci/mapd-charting/blob/9540c6a48993f97ecbf5b465cc1fc1e890ac74c2/src/mixins/raster-layer-poly-mixin.js#L603 . Looks like the filter interaction is broken in Choropleth chart. When we click on a polygon the selected polygon stayed colored but others are grayed out, and that is not working in the PR.

uyanga-gb:
Multilayer filter has known bugs, so not sure how we want to handle that in the View Chart Filters. I can see the eye icon on Master layer, but that will probably not work properly due to underlying bad bugs. Maybe best to hide the eye icon on Master layer tab?

uyanga-gb:
Also, I assume the filter updates dynamically in the popup while the map bounding box change. The map bounding box listener and the function that update the bounds are here https://github.com/omnisci/mapd-charting/blob/9540c6a48993f97ecbf5b465cc1fc1e890ac74c2/src/mixins/map-mixin.js#L274

thomasoniii:
I modified the original PR so that it doesn't change any of the existing mapd-charting behavior, it just exposes a new method to get at this underlying magic crossfilter filterObj. So the change is going to be non-destructive, at least. :-)

It does also all appear to be operating normally - the original issues about poly and zoom selection appear to have been backend related and resolved separately.

PR #378, opened 2019-10-14
[FE-8901] allow both fact and geo columns in custom choropleth cols

bmatcuk:
Solution utilizes a simplified SQL parser that determines what part of the custom sql column can be moved into the "fact" query (ie, the `WITH` part of the query), and the parts that cannot. The parser doesn't understand everything, but it does understand parentheses, functions (including aggregates), and math (including precedence). That's enough to cover most use cases for now.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [x] Fixes FE-8901

## :smoking: Smoke Test
- [x] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/raster-layer-poly-mixin.js
+++ b/src/mixins/raster-layer-poly-mixin.js
@@ -9,6 +9,7 @@ import d3 from "d3"
 import { events } from "../core/events"
 import { parser } from "../utils/utils"
 import { lastFilteredSize, setLastFilteredSize } from "../core/core-async"
+import parseFactsFromCustomSQL from "../utils/custom-sql-parser"
 
 const polyDefaultScaleColor = "rgba(214, 215, 214, 0.65)"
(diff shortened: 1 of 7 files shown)


src/mixins/raster-layer-poly-mixin.js
```diff
@@ -135,25 +137,37 @@ export default function rasterLayerPolyMixin(_layer) {
...
+        // Custom SQL may include references to both the base table and the
+        // geo-join table. The custom SQL is parsed to move references to the
+        // base table into the WITH clause, and everything else outside in the
+        // parent SELECT.
+        // eslint-disable-next-line no-extra-semi
+        ;({
+          factProjections: colorProjection,
+          factAliases: colorProjectionAs,
+          expression: colorField
+        } = parseFactsFromCustomSQL(
```
uyanga-gb: magic function ;)

PR #406, opened 2020-01-17
[FE-10291] Add range slider for bubble charts

CaitW:
[Required reading](https://github.com/omnisci/mapd-immerse/pull/6151)

See my [Immerse PR here](https://github.com/omnisci/mapd-immerse/pull/6169)

Herein lie changes that allow us to have a range slider for sized bubble charts in Immerse.

- Passes a boolean regarding whether or not the chart currently has a size measure
- Adds a default absolute minimum pixel radius for sized bubbles (bubble charts with a size measure)
- Adds a default minimum size ratio for sized bubbles
- Modifies `plotData` to have two different paths for assigning bubble size ranges, depending on whether there's a size measure or not. 
   - Unsized bubbles have a fixed pixel size 
   - Sized bubbles are scaled. This scale is a function of an arbitrary multiplier and the chart's width. This means that sized bubbles are responsive to changes in chart width. We allow the user to modify the multipliers for the lower and upper bounds. However, we have an *absolute* minimum radius of 2 pixels, so that bubbles don't disappear

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -33096,6 +33096,7 @@ function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { de
  */
 function bubbleMixin(_chart) {
   var _maxBubbleRelativeSize = 0.3;
+  var _minBubbleRelativeSize = 0.004;
 
   /* OVERRIDE ---------------------------------------------------------------- */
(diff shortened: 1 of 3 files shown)


CaitW:
retest this please

PR #409, opened 2020-01-24
[FE-10328] include capMixin for heatmap

uyanga-gb:
## Description: 
This issue has been present in the Heat chart for years. The infinite loop only happens when the chart is filtered (by clicking on a box or an axis) before the chart is saved. The click filter logic is in nested promises in mixins but that seems like assuming the cap-mixin.js logic is included like row and pie chart. Thus, without setting the cap value for the heat chart is causing the unresolved promise when the chart is initialized. Setting the cap value and including the cap-mixin in the heat chart doesn't seem to cause any issue. The default cap value is hardcoded because we don't have a slider to control it. Setting a reasonable cap value is essential to show enough boxes.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #https://jira.omnisci.com/browse/FE-10328

## Relates with: 
https://github.com/omnisci/mapd-immerse/pull/6196

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -9703,53 +9703,6 @@ function colorMixin(_chart) {
 Object.defineProperty(exports, "__esModule", {
   value: true
 });
-var events = exports.events = {
-  current: null
-
-  /**
-   * This function triggers a throttled event function with a specified delay (in milli-seconds).  Events
-   * that are triggered repetitively due to user interaction such brush dragging might flood the library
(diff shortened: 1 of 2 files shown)


uyanga-gb:
Looks like the issue is fixed already.

PR #414, opened 2020-02-21
strokeColor: "fillColor"

biovisualize:
Adding the special `strokeColor: "fillColor"` to make the stroke color teh same as the fill color

### DIFF
--- a/src/mixins/raster-layer-poly-mixin.js
+++ b/src/mixins/raster-layer-poly-mixin.js
@@ -486,6 +486,8 @@ export default function rasterLayerPolyMixin(_layer) {
       }
     }
 
+    const mark = state.mark || {}
+
     const marks = [
       {
         type: "polys",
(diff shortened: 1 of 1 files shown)


src/mixins/raster-layer-poly-mixin.js
```diff
@@ -500,14 +509,15 @@ export default function rasterLayerPolyMixin(_layer) {
...
           },
           fillColor,
+          /* 
+            "fillColor" is a special keyword to set strokeColor the same as fillColor
+            otherwise it will be strokeColor or white
+          */
           strokeColor:
-            typeof state.mark === "object" ? state.mark.strokeColor : "white",
-          strokeWidth:
-            typeof state.mark === "object" ? state.mark.strokeWidth : 0.5,
```
biovisualize: Notice that I changed that default 0.5 to be 0 as I don't understand why you would want to have a half pixel white border by default (I can understand the half pixel thing, it's usually for antialiasing, but not why you would want a white border by default). We default it to 0 in Immerse anyway so that's probably why we never noticed it. And we can make it 0.5 px in Immerse.

PR #437, opened 2020-06-12
[FE-11658] - hacked up mapd-table

thomasoniii:
Filters on non-grouped tables _do not work_.

This makes them work _better_.

You'd once again be able to click on a column to filter on it, including multiple filters. Clearing any one of those filters clears all the filters (as it always worked).

A table can now be loaded with a saved filter, and it will display "correctly" in that the data is filtered, but it will _not_ have the blue highlighting unless the filter is re-applied.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -60394,13 +60394,9 @@ function filterHandlerWithChartContext(_chart) {
       if (_chart.clearTableFilter) {
         _chart.clearTableFilter(); // global filter also will clear all the columns filters on the table
       }
-    } else if (_chart.hasOwnProperty("rangeFocused")) {
-      dimension.filterMulti(filters, _chart.rangeFocused(), _chart.filtersInverse(), _chart.group().binParams());
     } else if (_chart.getFilteredColumns && Object.keys(_chart.getFilteredColumns()).length > 0) {
       // case for column filtering on measures
       return filters;
-    } else {
-      dimension.filterMulti(filters, undefined, _chart.filtersInverse(), _chart.group().binParams()); // eslint-disable-line no-undefined
     }
     return filters;
   };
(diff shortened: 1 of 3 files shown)


dist/mapdc.js
```diff
@@ -60394,13 +60394,9 @@ function filterHandlerWithChartContext(_chart) {
       if (_chart.clearTableFilter) {
         _chart.clearTableFilter(); // global filter also will clear all the columns filters on the table
       }
-    } else if (_chart.hasOwnProperty("rangeFocused")) {
-      dimension.filterMulti(filters, _chart.rangeFocused(), _chart.filtersInverse(), _chart.group().binParams());
     } else if (_chart.getFilteredColumns && Object.keys(_chart.getFilteredColumns()).length > 0) {
```
thomasoniii, line 60397: `filterMulti` is now a no op, so just nuke these calls to it. These were the source of the _chart.group().binParams() does not exist errors.

dist/mapdc.js
```diff
@@ -84368,10 +84364,25 @@ function mapdTable(parent, chartGroup) {
...
+          var filterArray = [];
+          _chart.removeFilteredColumn(col.expression);
+          if (filterArray.some(function (f) {
+            return f !== undefined && f !== null;
+          })) {
+            _chart.onClick(filterArray); // will update global filter Clear icon
+          } else {
+            _chart.filterAll();
+          }
+          (0, _coreAsync.redrawAllAsync)(_chart.chartGroup());
```
thomasoniii, line 84379: In an ideal world, we'd populate the filterArray with just the filters that still exist, but it doesn't work. I don't want to make it work right now.

So instead, we just scrap all filters if you try to clear any of them. This is the same as 5.2's behavior.

dist/mapdc.js
```diff
@@ -84442,8 +84453,6 @@ function mapdTable(parent, chartGroup) {
     } else if (type === "DATE") {
       var dateFormat = _d2.default.time.format.utc("%Y-%m-%d");
       val = "DATE '" + dateFormat(val) + "'";
-    } else if (val && typeof val === "string") {
-      val = "'" + val.replace(/'/g, "''") + "'";
```
thomasoniii, line 84446: This was mapping `"foo"` -> `"'foo'"`. I have to assume this was done for sql building purposes and escaping things. We don't need it any more.

PR #451, opened 2020-07-23
[FE-11967] :bug: test for domains with cnt(NULL) > 1

bmatcuk:
PR #270 added a fix for domains that look like `["NULL", "NULL"]` - however, that fix was applying to any domain that included "NULL". This fix just tests for domains that have "NULL" more than once.

Fixes [FE-11967](https://omnisci.atlassian.net/browse/FE-11967)


bmatcuk:
retest this please

PR #454, opened 2020-08-10
[FE-12062][FE-12063] - logic fixes for shift to zoom

thomasoniii:
Comments inline below.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -34369,6 +34369,7 @@ function mapMixin(_chart, chartDivId, _mapboxgl) {
   // a way to re-enable it. Ideally, we'd just stop the event from firing the default actions, but
   // that doesn't seem to handle it. So here we are.
   function onMouseDownCheckForShiftToZoom(e) {
+    _map.boxZoom.disable();
     if (!e.originalEvent.shiftKey) {
       _map.scrollZoom.disable();
       _map.dragPan.disable();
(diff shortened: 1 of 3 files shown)


dist/mapdc.js
```diff
@@ -34369,6 +34369,7 @@ function mapMixin(_chart, chartDivId, _mapboxgl) {
   // a way to re-enable it. Ideally, we'd just stop the event from firing the default actions, but
   // that doesn't seem to handle it. So here we are.
   function onMouseDownCheckForShiftToZoom(e) {
+    _map.boxZoom.disable();
```
thomasoniii, line 34372: interacting with a lasso can re-enable the boxZoom filter, so we explicitly re-disable it when we click

dist/mapdc.js
```diff
@@ -79966,7 +79966,7 @@ var ScrollZoomHandler = function (_BaseHandler2) {
   }, {
     key: "_wheelZoom",
     value: function _wheelZoom(doFullRender, delta, e) {
-      if (this._chart.shiftToZoom() && e && !e.shiftKey) {
+      if (this._chart.shiftToZoom() && (!e || !e.shiftKey)) {
```
thomasoniii, line 79969: The logic here was slightly wrong - we can fire _wheelZoom events from a timeout which means there is no event attached.

So we want to bow out if it's shiftToZoom AND there's no event (it's the timeout - ignore it) OR we have an event but w/o the shift key.

PR #470, opened 2020-09-22
[FE-12306] Don't allow axis lock if x-domain is NaN

CaitW:
[See Immerse PR description](https://github.com/omnisci/mapd-immerse/pull/7108)

Occasionally, the x-axis domain of a chart can be empty due to a global or chart filter. mapd-charting will see the domain extent as `[NaN, NaN]`, and we don't want the user to lock the chart with this faulty extent. So, we're going to use this function to see if we should disable the axis lock feature. 

Conditions:

1. Only the x-axis lock toggle can be disabled (for now?)
2. We only want to prevent the user from locking the x-axis, not unlocking it
3. We want to disable it if any value in the extent === NaN

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


bmatcuk (approved):
lgtm

PR #476, opened 2020-10-19
Version bump

israelvicars:
Version bump for publishing changes made post-5.5 / pre-5.6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapd/mapdc",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mapd/mapdc",
-  "version": "1.1.0",
+  "version": "1.2.0",
```
israelvicars, line 3: Needs to be 2.0.0 to account for significant API changes

thomasoniii (approved):
just noting, 2.0.0 is to cover our butts in case the crossfilter removal changes broke anything noteworthy with the mods to mapd-charting.

PR #484, opened 2020-11-25
[FE-12664] Support custom column  in popup

uyanga-gb:
# Merge Checklist

## :wrench: Issue(s) fixed:
- [ ] Fixes #https://omnisci.atlassian.net/browse/FE-12664

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/charts/raster-chart.js
+++ b/src/charts/raster-chart.js
@@ -605,7 +605,7 @@ export default function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
       .then(results => callback(results[0]))
       .catch(error => {
         throw new Error(
-          `getResultRowForPixel failed with message: ${error.message}`
+          `getResultRowForPixel failed with message: ${error.error_msg}`
         )
       })
   }
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -88664,10 +88664,10 @@ function rasterLayer(layerType) {
   }
 
   _layer.displayPopup = function (chart, parentElem, result, minPopupArea, animate) {
-    // hit testing response includes color or size measures result as color or size
+    // hit testing response includes color or size measure's result as "color" or "size"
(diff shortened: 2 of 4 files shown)


src/charts/raster-chart.js
```diff
@@ -605,7 +605,7 @@ export default function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
       .then(results => callback(results[0]))
       .catch(error => {
         throw new Error(
-          `getResultRowForPixel failed with message: ${error.message}`
+          `getResultRowForPixel failed with message: ${error.error_msg}`
```
israelvicars, line 608: What's the reason for changing from the standard JS [Error.prototype.message](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message)? 

Not sure if it's related to this error since it's reportedly reproduced on master: https://github.com/omnisci/mapd-immerse/pull/7301#issuecomment-736739723

israelvicars (changes requested):

src/charts/raster-chart.js
uyanga-gb: yeah, this is not related with this PR. Tamara's error is not related with this. The error object doesn't have `message` property but `error_msg`. I am not sure when the response changed, but because of this we are not seeing the error message in the console.

uyanga-gb: The error she sees is the error that comes up when hit testing call fails. With this ^ change, we can see what the actual error message is.

israelvicars: 👍

israelvicars (approved):

PR #506, opened 2021-01-29
remove travis

bmatcuk:
Removes travis config and the script that builds the docs - that script has been recreated in jenkins, but is currently disabled since it hasn't been working since Apr 2019.


israelvicars (approved):
🚀

PR #519, opened 2021-03-15
[FE-13276] fix table formatting again

bmatcuk:
There's a mismatch between what Immerse uses to set the "keys" for formatting, and what mapd-charting is using as keys to get the formatting.


israelvicars (approved):
🚀 🚀 🚀

PR #522, opened 2021-03-25
[FE-11793] Copy from popup

uyanga-gb:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #https://omnisci.atlassian.net/browse/FE-11793

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


israelvicars (approved):
🚀

src/mixins/raster-layer.js
```diff
@@ -371,7 +373,17 @@ export default function rasterLayer(layerType) {
   }
 
   function renderPopupHTML(data, columnOrder, columnMap, formatMeasureValue) {
-    let html = ""
+    let html =
+      '<div class="' +
+      _popup_item_copy_class +
+      '">' +
+      '<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">' +
+      '<path d="M10.6668 0.666748H2.66683C1.9335 0.666748 1.3335 1.26675 1.3335 2.00008V11.3334H2.66683V2.00008H10.6668V0.666748ZM10.0002 3.33341L14.0002 7.33341V14.0001C14.0002 14.7334 13.4002 15.3334 12.6668 15.3334H5.32683C4.5935 15.3334 4.00016 14.7334 4.00016 14.0001L4.00683 4.66675C4.00683 3.93341 4.60016 3.33341 5.3335 3.33341H10.0002ZM9.3335 8.00008H13.0002L9.3335 4.33341V8.00008Z" fill="#0089D1"/>' +
```
biovisualize, line 381: We should remove the hardcoded `fill="#0089D1"` here and set it in css so we can override it, for example with user override in servers.json ($mdc-theme-primary) and with hover state. One trick is to add `fill="currentColor"` on the svg so you can override it with css with `.popup-item-copy {color: $mdc-theme-primary}` from Immerse.

biovisualize (approved):
I suggested an improvement. I will let you judge if we should do it now.

biovisualize (changes requested):
I just thought of something. Could you change the css to be 
```
.map-popup-box-new .popup-item-copy {
    visibility: visible;
    display: inline-block;
    position: absolute;
    top: 0;
    right: 0;
}

.map-popup-box-new {
    padding: 8px 12px 4px 8px;
}
```
It solves a few things:
* it gets rid of the space that it added on top
* it adds a little more right padding to make sure the icon doesn't overlap with the text
* only the icon is clickable, not an arbitrary empty space on top

It would be even better if we would have a click state because there's no indication that the copy-to-clipboard worked. But that's a stretch goal.

src/mixins/raster-layer.js
uyanga-gb: Removed the fill from svg and added the css in immerse.

uyanga-gb:
> I just thought of something. Could you change the css to be
> 
> ```
> .map-popup-box-new .popup-item-copy {
>     visibility: visible;
>     display: inline-block;
>     position: absolute;
>     top: 0;
>     right: 0;
> }
> 
> .map-popup-box-new {
>     padding: 8px 12px 4px 8px;
> }
> ```
> 
> It solves a few things:
> 
> * it gets rid of the space that it added on top
> * it adds a little more right padding to make sure the icon doesn't overlap with the text
> * only the icon is clickable, not an arbitrary empty space on top
> 
> It would be even better if we would have a click state because there's no indication that the copy-to-clipboard worked. But that's a stretch goal.

Made a change based on your suggestion @biovisualize. I also applied click state in immerse.

biovisualize (approved):
Nice!

PR #531, opened 2021-04-27
call elastic listeners after domain

bmatcuk:
When locking/unlocking the axis, call the elasticX/Y listeners after calling the domain listeners


israelvicars (approved):
:shipit:

PR #537, opened 2021-07-08
[FE-13791][IHS] - fix scatter chart related slowdowns

thomasoniii:
Scatter plot has a couple of stupid slowdowns, reported most recently by IHS.

First of all, it will always refresh the retrieved dataURL, even if it's identical to its previous value. The fix here is just to cache the img url + the render bounds, and not bother to do anything if they're both unchanged.

Secondly, during chart creation, it ends up firing off several `redrawAsync` calls before it has even completed its initial render. `setValueFormatter`, `setElasticX`, and `setElasticY` all trigger redraws.

The best way I could see to do this was to add new functionality to `async-mixin` - `enableChartRedraw`/`disableChartRedraw`. So that way we can disable it when we start to create a scatter plot, set the values, then re-enable it when we're done. It's a stupid hack, but so's everything else in mapdc. Note that due to the data involved and race conditions, extra queries are not always fired, but we usually end up with at least 1 extra.

To test, just create some scatter plots and confirm they worked as before, and watch the network tab to confirm that fewer/the same number of network requests fire off.

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -49955,8 +49955,20 @@ function asyncMixin(_chart) {
     });
   };
 
+  var _chartRedrawEnabled = true;
+  var chartRedrawEnabled = function chartRedrawEnabled() {
+    return _chartRedrawEnabled;
+  };
+
+  _chart.enableChartRedraw = function () {
+    _chartRedrawEnabled = true;
+  };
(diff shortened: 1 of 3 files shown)


bmatcuk (approved):
lgtm

PR #538, opened 2021-07-16
High precision lasso w/ st_contains

vastcharade:
Adds a whole new slew of changes/features to the lasso tool improving performance and precision.

Performance is improved by replacing the earcut triangulation filter set for each polygon shape w/ st_contains, now that st_contains is much improved as well as the recent additions of inline st_point construtors (enabling lon/lat columns to be constructed at query time as an ST_Point to be used in ST_Contains)

The ST_Contains approach had some caveats, namely the best settings for performance for st_contains is to do the operation in srid 4326 (WGS84) lon/lat coordinates. Unfortunately this means that straight lines in this space do not equate to straight lines in srid 900913 (web-mercator) space, which is the space the lasso polys are drawn. So, depending on the location and zoom of the view, there can be a noticeable gap between the drawn lasso and the returned results of the query using the st_contains filter thanks to the non-linear projection.

The goal then became to match the drawn poly via the lasso tool with the query results as much as possible. This would require having enough vertices to properly sample the draw lasso poly in WGS84 space. The question then is how much sampling is required? The answer is view dependent. You could be zoomed out to see the entire world, in which case you may not need any sampling, or zoomed in really tight at near 1m resolution, in which case you may need heavy sampling. So the algorithm is largely a dynamic level-of-detail algorithm, that auto-subdivides the segments of the drawn lasso poly depending on it's scale/positioning relative to the current view. The basic idea is If the line segment intersects the current view and the length of that intersection is > 40 pixels, it should be subdivided w/ a max of 40 pixel subsegments (the 40 pixel threshold was picked arbitrarily after some empirical testing). The algorithm ends up working well, and dynamically subdivides only line segments in view, limiting the total number of subdivided points and keeping performance reasonable. The resulting draw of the polygon matches the results of ST_Contains exactly.

I then took this a step further and wanted to do the same for the circle lasso tool. The circle lasso tool uses the `distance_in_meters` extension function for its filter operator. `distance_in_meters` uses the [haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) for calculating distance between two WGS84 lon/lat points. So the circle here has similar issues as the lasso poly above: a drawn circle in web-mercator does not equate to a circle in WGS84. The circle tool has long countered this issue by doing a custom circle draw by drawing line segments between 60 vertices rotated about a center point in WGS84 coordinates. However, those 60 line segments can be prone to the same gap problem as the ST_Contains lasso depending on zoom/position. So in a general sense, these line segments could benefit from a similar auto-subdivide algorithm as the lasso poly. However, there are many differences, the main one being that in this case the line segment actually represents a fragment of the full 360 arc of the circle, so there were a large number of different edge cases to handle for the circle case compared to the lasso poly. I will not cover those details here, but am happy to answer any questions. In any event, the result of this algorithm is equivalent to the poly, a dynamic, auto-subdivide algorithm that only impacts line segments/circle arcs in view (or near view), limiting the number of subdivided points drawn, and it matches the results of the `distance_in_meters` filter exactly.

**SUMMARY**

 * Replaces the earcut triangulation filter of lasso polys with ST_Contains for both map charts (which uses srid 4326) and scatterplot charts (which uses srid 0)
 * Improves the precision of all the math used w/ the lasso tool to 64-bits, which means you can draw circles or line segments at full world scale, or at 1cm or less scale and everything just works - no numerical instability.
 * Adds a new LatLonPoly class that draws lasso polys drawn on maps which auto-subdivides line segments in view in order to match the results of its new ST_Contains filter op.
 * Significantly improves the precision of the previously existing LatLonCircle class by not only taking advantage of the 64-bit precision, but also dynamically auto-subdividing the custom line-segmented circle in order to maintain the look of a circle at any zoom level and match the results of the `distance_in_meters` filter op.

**NOTE for reviewers:**

I was not able to build latest master of the mapd-charting library locally on my ubuntu 20.04 workstation. I had to change some of the dependencies/settings to get it to work. This PR does not include those changes (if you're curious, you can look at the [commits prefixed w/ "(CROOT only)" in my chris/lasso_tool_improvements_1b branch](https://github.com/omnisci/mapd-charting/commits/chris/lasso_tool_improvements_1b)

As such, this PR does not include the right `package-lock.json` nor `dist/mapdc.js` files. They would need to be regenerated and appended to this PR. How should we handle that? Perhaps someone on the FE team could checkout this PR branch and regenerate those and add the results as a new commit? Probably the best way.

And lastly, I apologize for dumping this massive PR on you at the end-of-day on a friday. I do not expect anyone to thoroughly vet this code manually. So perhaps it may be a better use of everyone's time to get together in a 30-60 minute meeting where I can summarize the code. I'll leave that to the FE powers that be :) But to be honest it's really not all that huge. A big chunk of the resulting code is comments as well as verbose operator/variable names that get translated into a multi-line call by the code formatter when in fact it is a single function call. The net effect is that it looks a lot bigger than it actually is.

Anyway, the goal is for this to get into 5.7. I hope that is still a possibility.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


biovisualize:
Retest this please

israelvicars (approved):
Merging fixed build for 5.7 RC testing

PR #540, opened 2021-07-22
[FE-13887] table param coupling

thomasoniii:
I set up [https://omnisci.atlassian.net/browse/FE-13887](FE-13887) as a centralized ticket for these param/table issues. There are probably more related tickets.

This adds in a hook to mapd-table to allow us to process columns with parameters when they're pulled out for filtering purposes, as opposed to pre-processing them and storing that string on the dimension. Immerse sets the `setCustomRetrieveFilterType` function.

That, in turn, allows us to track usage only when data is fetched.

Also - mapd-table blew up when you attempted to filter on any measure columns that weren't in the table. So it's not just that `${foo}` would break, things like `avg(foo)` or `8` would also break. So this tweaks it to refuse to set a filter entirely if it's not in the table.

There may be cases where you could have set a valid filter with a column of that type, and those would need to be addressed. If such a case exists, then previous to this change it would've crashed immerse. After this change, it will merely prevent creation of that filter (which would've crashed immerse). If such a case can be proven, we should amend the callback in Immerse to allow passage of a type value to allow a filter to be created.

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -85392,6 +85392,8 @@ function mapdTable(parent, chartGroup) {
     });
   }
 
+  // our default retriever just takes the expr (column), sticks the column on it, then snags
+  // the type from the columns list. Being careful not to blow up if it doesn't exist.
   var _retrieveFilterColType = function _retrieveFilterColType(_ref) {
     var expr = _ref.expr,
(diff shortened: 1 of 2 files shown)


bmatcuk (approved):
:shipit:

PR #541, opened 2021-08-04
[FE-13941][others?] params and poly stuff

thomasoniii:
This is a hack that I _think_ will keep us in action but it should be re-visited and done better in the nebulous future.

The problem was correctly diagnosed - the params were preprocessed before hitting the query, so usage couldn't be noted. We needed to hold onto the params until the end for the notation. However, if we did that, then it blew up at an earlier step. 

This is because custom colors are parsing out the query and building an AST and I _swear_ it's doing something with its own templating at this level, but I haven't isolated or proven that.

So this mod here just lets us override some behavior - provide a function to retrieve the color aggregate, and another function to post-process the values that come out of `parseFactsFromCustomSQL`.

Those functions are handed in via immerse and discussed in that PR.

Longer term, `parseFactsFromCustomSQL` and/or `buildAst` could be modified to deal with parameters directly, or we could move notation to another location...but just for the custom color query, which'd suck as a special case.

The immediate ramification is that I _think_ it still works properly, but is likely to break if you have a custom color measure that contains columns from the color table, fact table, and a parameter, since it just blindly re-inserts the original query if a param is found. We can fix that other bug if it's ever reported, possibly by moving notation around and pre-processing, but I really don't want do that.

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -55818,6 +55818,7 @@ function rasterLayerPolyMixin(_layer) {
         colorField = _parseFactsFromCustom.expression;
       }
 
+      ;
       var _customColorProjectio = _customColorProjectionPostProcessor(color.aggregate, {
         colorProjection: colorProjection,
         colorProjectionAs: colorProjectionAs,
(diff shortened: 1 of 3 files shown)


sabrichu (approved):
Also seems not crazy

thomasoniii:
retest this please

PR #543, opened 2021-08-09
[FE-13942] - table redraw change

thomasoniii:
I dunno if this is really worth incorporating at this point for 5.7. It's a trivial change, but it's table chart, so it's _never_ a trivial change.

Anyway, way back in the olden days of _four years ago_, mapd-table was wired up so that when you click the sort button, it'd fire off a `redrawAllAsync` on the chart's group, which'd cause needless re-querying on other charts.

This changes to just call the single chart's `redrawAsync` in its place. I think it works, but I'll never have confidence in anything table related. 

And it looks like an ancient issue. So maybe it's not worth sticking in in the last minutes of 5.7? I dunno.

### DIFF
--- a/dist/mapdc.js
+++ b/dist/mapdc.js
@@ -85418,7 +85418,7 @@ function mapdTable(parent, chartGroup) {
         }
 
         _chart._invokeSortListener(_sortColumn);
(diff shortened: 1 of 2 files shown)


bmatcuk (approved):

israelvicars:
Let's roll the dice 🎲

PR #547, opened 2021-09-02
[FE-13762] :bug: fix legend getting disconnected from dom

bmatcuk:
No idea why it happens, but sometimes the legend either seems to get completely disconnected from the DOM, or the legend in the DOM doesn't match the class instance.

¯\\\_(ツ)_/¯


RavenHursT:
This looks like it fixes the edge case I was seeing w/ the canceled chart's legend not updating when changing a param value.. so.. go for it. #YOLO

RavenHursT (approved):

PR #550, opened 2021-09-23
[FE-14191][FE-14196] Limit pretty-printed aliases to data exports

cjvil:
Only pretty-print aliases when exporting data. Custom SQL dimensions could break the generated query here, and hit-testing also seems to depend on the previous `key[#]` aliases.

(This does _not_ fix exporting data with custom SQL dimensions—wrapping the group by in quotation marks seems to generate valid SQL, but still fails validation somewhere down the line?)

Fixes FE-14191, FE-14196


bmatcuk (approved):
lgtm

PR #560, opened 2022-02-03
[FE-14661] Replaces distance_in_meters with st_dwithin for lasso circle filters

vastcharade:
Also changed the airplane symbol mark to be normal "symbol" since I was in here. The airplane was added to our stock symbols for the `symbol` mark some time ago.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes [FE-14661]

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

[FE-14661]: https://omnisci.atlassian.net/browse/FE-14661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mixins/raster-draw-mixin.js
+++ b/src/mixins/raster-draw-mixin.js
@@ -235,7 +235,7 @@ export function rasterDrawMixin(chart) {
                 LatLonUtils.conv900913To4326(pos, pos)
                 const meters = shape.radius * 1000
                 filterObj.shapeFilters.push(
-                  `DISTANCE_IN_METERS(${pos[0]}, ${pos[1]}, ${px}, ${py}) < ${meters}`
+                  `ST_DWithin(CAST(ST_SetSRID(ST_Point(${pos[0]}, ${pos[1]}), 4326) as GEOGRAPHY), CAST(ST_SetSRID(ST_Point(${px}, ${py}), 4326) as GEOGRAPHY), ${meters})`
                 )
               } else if (shape instanceof MapdDraw.Circle) {
                 const radsqr = Math.pow(shape.radius, 2)
--- a/src/mixins/raster-layer-point-mixin.js
+++ b/src/mixins/raster-layer-point-mixin.js
@@ -581,7 +581,7 @@ export default function rasterLayerPointMixin(_layer) {
 
     const marks = [
       {
-        type: markType === "airplane" ? "legacysymbol" : "symbol",
+        type: "symbol",
         from: {
           data: layerName
         },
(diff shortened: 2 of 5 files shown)


vastcharade:
@israelvicars Added you as a reviewer since I didn't know who else to add. Feel free to forward on to anybody else you deem more appropriate to review.

@tmostak You may want to double-check that the ST_DWithin expression is what you'd expect since I'm not that familiar with all the sql semantics here. I was just going off what you had pasted in slack a couple months ago: https://omnisci.slack.com/archives/C6HHEF0Q6/p1639079073152300

src/mixins/raster-draw-mixin.js
```diff
@@ -235,7 +235,7 @@ export function rasterDrawMixin(chart) {
                 LatLonUtils.conv900913To4326(pos, pos)
                 const meters = shape.radius * 1000
                 filterObj.shapeFilters.push(
-                  `DISTANCE_IN_METERS(${pos[0]}, ${pos[1]}, ${px}, ${py}) < ${meters}`
+                  `ST_DWithin(CAST(ST_SetSRID(ST_Point(${pos[0]}, ${pos[1]}), 4326) as GEOGRAPHY), CAST(ST_SetSRID(ST_Point(${px}, ${py}), 4326) as GEOGRAPHY), ${meters})`
```
vastcharade, line 238: NOTE: this always uses an ST_Point constructor, even if the input column is already a POINT type. This is because the px/py is already separated out to be ST_X(point), ST_Y(point) respectively by the time we reach here. To avoid the unnecessary cast would require unwrangling a bunch of upstream code that looks like it could be a bit of a pain, so I decided to punt that detail.

src/mixins/raster-layer-point-mixin.js
```diff
@@ -581,7 +581,7 @@ export default function rasterLayerPointMixin(_layer) {
 
     const marks = [
       {
-        type: markType === "airplane" ? "legacysymbol" : "symbol",
+        type: "symbol",
```
vastcharade, line 584: airplane symbol can use the normal `symbol` type now. that was added with this PR: https://github.com/omnisci/omniscidb-internal/pull/5660

israelvicars (approved):
:shipit: 

Thanks for the notes! That'll be helpful for reference.

vastcharade:
Thanks @israelvicars . I take it you will handle getting this over into immerse then too?

PR #567, opened 2022-03-25
[FE-15044] first draft adding align buttons to table chart

bmanturner:
This adds functionality to the table chart that allows the user to specify content alignment on a per-column basis. The buttons appear when you hover over the column, and can be accessed/manipulated via `chart.columnAlignments()`

Work will be done in Immerse to persist/populate this settings in the dashboard state.

[JIRA](https://heavyai.atlassian.net/browse/FE-15044)

CleanShot 2022-03-25 at 12 43 33@2x


bmatcuk (approved):
lgtm

PR #576, opened 2022-05-04
[FE-15240] Color ramp prototype

cmatzenbach:

Add necessary updates to charting to process and handle vega spec changes needed for color ramps.

Closes [FE-15240](https://heavyai.atlassian.net/browse/FE-15240)

### :mag: Verification Steps

1. Need to test through immerse - add all color ramps and verify map re-renders and vega query changed to using `threshold` when a complete set of color ramps is passed through. (Can use branch `matzy/FE-15240_color-ramp-prototype` if want to test locally in immerse)

### DIFF
--- a/src/utils/utils-vega.js
+++ b/src/utils/utils-vega.js
@@ -31,9 +31,7 @@ export function adjustRGBAOpacity(rgba, opacity) {
 }
 
 export function parseColorRamps(cr) {
-  console.log(cr)
-  console.log(typeof cr)
-  if (cr) {
+  if (cr && typeof cr === "object") {
     const cr_arr = cr
       .flatMap(a => a.filter(b => b !== "min" && b !== "max" && b !== ""))
       .map(a => parseFloat(a))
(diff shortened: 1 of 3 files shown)


src/utils/utils-vega.js
```diff
@@ -709,7 +721,35 @@ export function getScales(
...
+          return adjustRGBAOpacity(c, interp)
+        }),
+      accumulator: "density",
+      minDensityCnt: "-2ndStdDev",
+      maxDensityCnt: "2ndStdDev",
+      clamp: true
+    })
+  }
+
+  if (
```
bmatcuk, line 748: Might be more clear to write this something like:

```js
if (typeof color  === "object" && color.type === "density") {
  const parsedColorRamps = parseColorRamps(colorRamps)
  const domain = parsedColorRamps.length === color.range.length - 1
    ? parsedColorRamps
    : // original domain code
  // original code, but using the domain variable
}
```

Saves you from parsing the color ramps multiple times and duplicating a bunch of code.

bmatcuk (changes requested):
Looks alright, but you need to also run `npm run build` and commit the built version with your PR.

src/utils/utils-vega.js
cmatzenbach: Fixed - great catch

bmatcuk (approved):
👍

bmatcuk (changes requested):
Oh, wait! You still need to run `npm run build` and commit the built file with the PR

cmatzenbach:
@bmatcuk Ah crap, almost forgot.  Just committed it!

bmatcuk (approved):
lgtm

PR #577, opened 2022-05-31
[FE-15434] max bounds to map

thomasoniii:
Let's you specify the maxBounds on your map, and you _cannot_ zoom out beyond them.

This is useful if there is a country which does not exist and that you don't want people to see.


bmatcuk (approved):
👍

PR #585, opened 2022-09-30
Adds windbarb raster chart layer via a new raster layer state parser/validator/vega-generator

vastcharade:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


package.json
```diff
@@ -70,12 +70,13 @@
     "wellknown": "^0.5.0"
   },
   "devDependencies": {
-    "@heavyai/connector": "omnisci/mapd-connector#d75148077a2a4b23d66d89c63d05c20e5735af2e",
-    "@heavyai/crossfilter": "omnisci/mapd-crossfilter#6f291ac",
+    "@heavyai/connector": "https://github.com/heavyai/heavyai-connector.git",
+    "@heavyai/crossfilter": "https://github.com/heavyai/heavyai-crossfilter.git",
```
RavenHursT, line 74: We need to insure that we retain the commit hash references on these.  We've never gotten around to utilizing actual release numbers for our dependencies... 😕

vastcharade: ok - what I was really trying to do here was update the repo url to the new heavyai path.. I'll add the commit hash too.

vastcharade: so the `d751480` for heavyai-connector is ancient -- over 3 years old. Is there any reason to stay on that commit hash? I presume we should just bump it to the latest.

Similar for crossfilter - the `6f291ac` commit is over 2.5 yrs old.

RavenHursT: Oh wow.. we haven't updated that dependency in that long? lol.. figures...

Looks like this is the current hash we have immerse pointed at for connector: https://github.com/heavyai/heavyai-connector/tree/a2d43593c9fa9893919c0fe955c64ed14cd8262b

As for crossfilter.. We ripped `crossfilter` out of immerse ages ago... I don't see an `import` for `crossfilter` anywhere, so we _should_ be able to just remove that dep 🤞

RavenHursT: in fact.. now that I check.. I don't see any `import`s for `connector` either.. so we can prolly just rip that one out of the deps as well... 🤷

vastcharade: both are definitely dependencies at the charting level.

RavenHursT: They're in the `package.json`, for sure.. but I don't see anything in the project `import`ing them...

RavenHursT: Actually.. looks like they're only ever imported into your example file: https://github.com/heavyai/heavyai-charting/pull/585/files#diff-2affdc53e5d92c93e5e4e51405947ea6381a8fe18aef629ce21a1bcccddb6eacR2-R3

vastcharade: my guess is historically we've relied on the package being available in the global namespace and didn't use proper import syntax. we should obviously change that. but they're definitely dependencies.

src/mixins/raster-layer-windbarb-mixin.js
```diff
@@ -4241,8 +4241,8 @@ export default function rasterLayerWindBarbMixin(_layer) {
   //     chart.pushLayer("pointmap", _layer);
   //   }
   // };
-  // _layer.getLayerNames = function(chart) {
-  //   return chart.getLayerNames();
-  // };
+  _layer.getLayerNames = function(chart) {
```
vastcharade, line 4244: On the surface it feels like the caller would need to call this method like so:

```
layer.getLayerNames(chart)
```
but that seems superfluous. Why not just call the `getLayerNames()` method directly from `chart` in that case: 

```
chart.getLayerNames()
```

vastcharade:
retest this please

vastcharade:
retest this please

israelvicars (approved):
🚀

PR #589, opened 2022-11-02
Implement contour charts

johallar:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


example/exampleContour.html
```diff
@@ -0,0 +1,210 @@
...
+          type: "lines",
+          lineJoin: "bevel",
+        },
+        encoding: {
+          // Here so that lines layer doesn't break
+          size: "auto",
+          color: {
+          },
+          // Contour stuff
+          stroke: {
```
vastcharade, line 109: curious, was the `stroke` property exposed/working before w/ scales? I'm guessing not. I don't think there's a current immerse case where a scale is used for stroke width (except for maybe th auto-sizing thing, which I can't claim to know how that works)

I think this is fine for now, but I think it would be worth revisiting this to avoid doing a property specialization inside the encoding section. For instance, I see you can attach a color to the major/minor here, but how does that work with the `color` property above? What happens if there's a `color` defined above, and a `color` here for major/minor?

I think we should instead look at something like this, where we define an attribute name/tag or major/minor in the `contour` data block, so something like this:

```
{
  ...,
  data: [
    {
      type: "contour",
      ...,
      intervals: {
        major: config.major_contour_interval,
        minor: config.minor_contour_interval,
        columnName: "is_major_contour" // output column name that tags major/minor
      },
      ...
    }
  ],
  ...,
  encoding: {
    ...,
    color: {
      field: "is_major_contour",
      type: "nominal",
      scale: {
        domain: [true, false],
        range: ["red", "blue"]
      }
    },
    strokeWidth: {
      field: "is_major_contour",
      type: "nominal",
      scale: {
        domain: [true, false],
        range: [3, 1]
      }
    },
    ...
  }
}
```

What I like about this approach is the `color` and `strokeWidth` properties have no specializations for major/minor, and going one step further, we could use styles to avoid redundancy.

However, you do lose some of the explicit-ness/readability of your json.

src/mixins/map-mixin.js
```diff
@@ -439,7 +439,7 @@ export default function mapMixin(
           resetRedrawStack()
           console.log("on move event redrawall error:", error)
         })
-      } else if (_viewBoxDim !== null && layer.getState().data.length < 2) {
+      } else if (_viewBoxDim !== null && typeof layer !== "undefined" && layer.getState().data.length < 2) {
```
vastcharade, line 442: i assume this means you hit a case where `layer` was undefined somewhere, resulting in an error? do you happen to recall the exact case?

vastcharade (approved):
lgtm. this is a good approach to handling the contour specialization in general.

I think the contour definition at the vega-lite layer could use another round of revision to avoid potential confusion between the contour specializations in the `encoding` and pre-existing attributes (like `color` & `strokeWidth`), but this is definitely a good starting place, and worth landing to unblock some immerse work.

johallar:
retest this please

src/mixins/map-mixin.js
johallar: I think this was happening any time you moved the map around after the initial load... having a hard time testing with kali/mahakali being down but will look back into this once i'm up and running again

johallar: Testing today, this doesn't seem to be a problem anymore, seems like it must have been a result of some buggy contour code earlier resulting in an undefined layer. I'm going to revert!

example/exampleContour.html
johallar: Im going to revisit this when doing the contour polygon changes and leave it this way for now, but I agree with the suggestion and will shoot for a format similar to this on the next go

PR #592, opened 2022-11-09
Contour polygon layer

johallar:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/charts/raster-chart.js
+++ b/src/charts/raster-chart.js
@@ -351,14 +351,18 @@ export default function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
     const mapBounds = chart.map().getBounds()
     const geoTable = _layer.getState().encoding.geoTable
     const geoCol = _layer.getState().encoding.geocol
+    const data = _layer.getState().data
 
     let preflightQuery
     if (geoTable && geoCol) {
       preflightQuery = `SELECT COUNT(*) AS n FROM ${geoTable} WHERE ST_XMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lng} AND ST_XMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lng} AND ST_YMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lat} AND ST_YMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lat}`
-    } else {
+    } else if (data && data.length && data[0].table){
       preflightQuery = `SELECT COUNT(*) AS n FROM ${
         _layer.getState().data[0].table
       }`
+    }
+    if (!preflightQuery) {
+      return Promise.reject("Error generating preflight query, 'data' is empty and `geoTable` and `geoCol` are not specified.")
     }
     return chart.con().queryAsync(preflightQuery, {})
   }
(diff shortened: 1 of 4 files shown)


src/mixins/raster-layer-poly-mixin.js
```diff
@@ -624,33 +690,38 @@ export default function rasterLayerPolyMixin(_layer) {
...
-    const otherChartFilters = allFilters.filter(
-      (f, i) =>
-        i !== _layer.dimension().getDimensionIndex() && f !== "" && f != null
-    )
-
+  _layer._genVega = function(chart, layerName) {
     let polyFilterString = ""
-    let firstElem = true
+    let bboxFilter = ""
+    if (!isContourType(_layer.getState())) {
```
johallar, line 696: Dramatic diff... this just doesn't do the bounding box stuff if its a contour chart, wraps it in `if (!isContourType(_layer.getState())) {`

johallar: Also separated some logic into `getPolygonScale` a separate function above here

src/charts/raster-chart.js
```diff
@@ -352,8 +352,14 @@ export default function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
     const geoTable = _layer.getState().encoding.geoTable
     const geoCol = _layer.getState().encoding.geocol
 
-    const preflightQuery = `SELECT COUNT(*) AS n FROM ${geoTable} WHERE ST_XMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lng} AND ST_XMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lng} AND ST_YMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lat} AND ST_YMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lat}`
-
+    let preflightQuery
+    if (geoTable && geoCol) {
+      preflightQuery = `SELECT COUNT(*) AS n FROM ${geoTable} WHERE ST_XMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lng} AND ST_XMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lng} AND ST_YMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lat} AND ST_YMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lat}`
+    } else {
+      preflightQuery = `SELECT COUNT(*) AS n FROM ${
```
vastcharade, line 359: I wonder if this should do an extra level of validation -- seems a bit presumptuous for it to assume that state.data[0].table exists. So, wondering if this should check that `state.data.length === 1` and that `state.data[0].type === "contour"` before setting the preflight. Then, at the end of the method, it would throw an error if `preflightQuery` is empty.

src/mixins/raster-layer-poly-mixin.js
```diff
@@ -495,14 +427,9 @@ export default function rasterLayerPolyMixin(_layer) {
         ),
         default: adjustOpacity(
           polyDefaultScaleColor,
-          state.encoding.color.hasOwnProperty("showOther") &&
-            state.encoding.color.showOther === false
-            ? 0
-            : state.encoding.color.opacity
-            ? state.encoding.color.opacity
-            : 0.65
+          hasShowOther ? 0 : state.encoding.color.opacity || 0.65
```
vastcharade, line 430: the `0.65` magic number should be put in a global default variable for opacity. Obviously that predates you, so no need to do anything about it here. Just noting.

vastcharade (approved):
lgtm. Looks like there's all previous logic is left intact while layering in your changes. Noting one small place to validate incoming data, but otherwise looks good to me.

src/mixins/raster-layer-poly-mixin.js
johallar: Easy to update, i'll make the change along with the extra validation changes

src/charts/raster-chart.js
johallar: Returning rejected promise if fails this check now

PR #593, opened 2022-11-10
Reformat contour state

johallar:
⚠️ This is built on top of the polygons stuff, so I set the target branch to that for diff's sake. Will merge this after that one and rebase on master

This also includes popups in the contour demo and some small changes around that

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


scss/chart.scss
```diff
@@ -1050,7 +1050,7 @@ body {
         }
     }
 
-    .map-polyline{
+    .map-polyline g {
```
johallar, line 1053: This was overriding inline styles for the `<path..` so I made this more specific so that inline styles actually work when passed through `popupStyle` getter/setter

src/charts/raster-chart.js
```diff
@@ -334,8 +334,12 @@ export default function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
     const geoTable = _layer.getState().encoding.geoTable
     const geoCol = _layer.getState().encoding.geocol
 
-    const preflightQuery = `SELECT COUNT(*) AS n FROM ${geoTable} WHERE ST_XMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lng} AND ST_XMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lng} AND ST_YMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lat} AND ST_YMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lat}`
-
+    let preflightQuery;
+    if (geoTable && geoCol) {
+      preflightQuery = `SELECT COUNT(*) AS n FROM ${geoTable} WHERE ST_XMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lng} AND ST_XMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lng} AND ST_YMax(${geoTable}.${geoCol}) >= ${mapBounds._sw.lat} AND ST_YMin(${geoTable}.${geoCol}) <= ${mapBounds._ne.lat}`
+    } else {
+      preflightQuery = `SELECT COUNT(*) AS n FROM ${_layer.getState().data[0].table}`
```
johallar, line 341: This doesn't work for contours, which has points in and geocol out, just gets a total count in this case.

src/mixins/raster-layer-poly-mixin.js
```diff
@@ -565,17 +494,153 @@ export default function rasterLayerPolyMixin(_layer) {
...
+    filtersInverse,
+    layerName,
+    useProjection
+  }) {
+    const autocolors = usesAutoColors()
+    const getStatsLayerName = () => layerName + "_stats"
+    const state = _layer.getState()
+
+    let data;
+    if (isContourType(state)) {
```
johallar, line 530: Very dramatic diff for some reason, this check was what I added `if (isContourType(state))`

src/mixins/raster-layer-poly-mixin.js
```diff
@@ -398,88 +399,17 @@ export default function rasterLayerPolyMixin(_layer) {
...
-            expr: "min(maxcol, avgcol+2*stdcol)",
-            as: "maxcolor"
-          }
-        ]
-      })
-    }
-
-    const scales = []
-    let fillColor = {}
+  function getPolygonScale({state, layerFilter, layerName, autocolors}) {
```
johallar, line 405: Dramatic diff part 2: also moved this to an individual function

example/exampleContour.html
```diff
@@ -145,51 +145,40 @@
...
-              color: "#666666",
-              opacity: 0.75
-            },
-            major: {
-              width: 2,
-              color: "#666666",
-              opacity: 0.8
+          geocol: "contour_lines",
+          color: {
+            field: "is_major",
```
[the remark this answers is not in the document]
vastcharade: oh, looking further below in the `buildContourSQL` method, it seems you are saying this "field" value will define the name of that column. That certainly does make things less redundant, but what happens if the `field` is misnamed/different in the `strokeWidth`/`opacity`?

Anyway, I like this field approach to reduce redundancy, but we would need a plan to coordinate with the other encodings that might want to make use of the same `is_major` field.

src/utils/utils-contour.js
```diff
@@ -61,7 +65,7 @@ export const buildContourSQL = (state, mapBounds, isPolygons = false) => {
   const contourParamsSQL = buildParamsSQL(contourParams)
 
   const geometryColumn = isPolygons ? "contour_polygons" : "contour_lines"
-  const contourLineCase = `CASE mod(cast(contour_values as int), ${major_contour_interval}) WHEN 0 THEN 1 ELSE 0 END as is_major `
+  const contourLineCase = `CASE mod(cast(contour_values as int), ${major_contour_interval}) WHEN 0 THEN 1 ELSE 0 END as ${encoding.color.field} `
```
vastcharade, line 68: utlimately I feel like including the CASE and such should be optional depending on whether the user is wants major/minor interval distinction or not.

vastcharade, line 68: Also, what happens if `encoding.color` doesn't exist? i.e. the user wants to use a flat color for all the contours, but wants them sized differently based on major/minor interval - so encoding.strokeWidth.field exists but not color.

Also, what happens if the user wants to color the contours by some other potential measure, like the length of the line? -- not sure we would have that kind of control initially, but it's possible a user may want to color / size by something other than the is_major distinction at some point. We would need to think about handling that.

vastcharade (changes requested):
Overall seems fine, tho I think we should review the way the `is_major` column name is selected and utilized.

example/exampleContour.html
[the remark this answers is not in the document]
johallar: Yeah wasn't quite sure how to handle that, maybe just move it to a root property of encoding, like `encoding.majorMinorField`? Or are there any existing examples of something like this?

example/exampleContour.html
```diff
@@ -145,51 +145,40 @@
...
+            type: "nominal",
+            scale: { 
+              domain: [0, 1],
```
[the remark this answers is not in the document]
johallar: I'll give this a try

example/exampleContour.html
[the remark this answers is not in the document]
johallar: Oh you mentioned putting it in the data block, i'll see what this looks like!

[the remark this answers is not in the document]
johallar: Hm so I think it still makes sense to set it up like this:
```
const majorMinorField = "is_major"
...
contourLayer.setState({
        data: [{
          type: "contour",
          majorMinorField,
          ...
        }],
        encoding: {
          color: {
             field: majorMinorField,
             ...
          },
          ...
        }
    })
```

This leaves it open in the future if we wanted to setup other things to drive color/strokeWidth/opacity, but charting will only get the field name from one place. Lemme know if that makes sense to you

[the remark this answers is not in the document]
vastcharade: yes, I was going to suggest something similar. It might also make sense to namespace the major/minor-related properties in a sub-block, so:

```
{
  "type": "contour",
  ...,
  "intervals": {
    "major": <major interval>,
    "minor": <minor interval>,
    "field_name": "is_major"
  },
  ...
}
```

example/exampleContour.html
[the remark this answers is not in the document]
johallar: This works! I'll make the change
```
{
  "width": $width,
  "height": $height,
  "viewRenderOptions": { "premultipliedAlpha": false },
  "data": [
    {
      "name": "contourLayer",
      "format": "lines",
      "sql": "select \n    contour_lines, \n    contour_values,\n    CASE mod(cast(contour_values as int), 500) WHEN 0 THEN 1 ELSE 0 END as is_major \n  from table(\n    tf_raster_contour_lines(\n      raster => cursor(select raster_lon, raster_lat,  z from colorado_dem), contour_interval => 100.0, agg_type => 'AVG', fill_agg_type => 'AVG', bin_dim_meters => 90.0, contour_offset => 0.0, neighborhood_fill_radius => 0, fill_only_nulls => FALSE, flip_latitude => FALSE\n    )\n  )",
      "enableHitTesting": false
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "linear",
      "domain": [-14086369.117126567, 3127822.7956852647],
      "range": "width",
      "nullValue": -100
    },
    {
      "name": "y",
      "type": "linear",
      "domain": [-556431.0937679282, 4972432.628235508],
      "range": "height",
      "nullValue": -100
    },
    {
      "name": "contour_width",
      "type": "ordinal",
      "domain": [false, true],
      "range": [1, 2]
    },
    {
      "name": "contour_color",
      "type": "ordinal",
      "domain": [false, true],
      "range": ["rgba(153,0, 0,0.75)", "rgba(40,0,100,0.8)"]
    }
  ],
  "projections": [
    {
      "name": "mercator_map_projection",
      "type": "mercator",
      "bounds": {
        "x": [
          $minLon,
          $maxLon
        ],
        "y": [
          $minLat,
          $maxLat
        ]
      }
    }
  ],
  "marks": [
    {
      "type": "lines",
      "from": { "data": "contourLayer" },
      "properties": {
        "x": { "field": "x" },
        "y": { "field": "y" },
        "strokeColor": { "scale": "contour_color", "field": "is_major" },
        "strokeWidth": { "scale": "contour_width", "field": "is_major" },
        "lineJoin": "bevel"
      },
      "transform": { "projection": "mercator_map_projection" }
    }
  ]
}
```

src/utils/utils-contour.js
johallar: If you're OK w it i'm going to leave the CASE in there for now. This handles making distinctions in the styles and not, albeit not as efficiently if the user makes major/minor settings the same.

example/exampleContour.html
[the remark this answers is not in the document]
johallar: maybe `intervalTypeFieldName`? Can't go wrong with super descriptive?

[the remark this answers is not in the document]
johallar: Updated the format to match above with the descriptive field name property, if you prefer `field` i'm happy to change it back

src/utils/utils-contour.js
johallar: This should be handled by formatting changes above, this now reads from `data.interval.intervalTypeFieldName`, but let me know if you're looking for something else

example/exampleContour.html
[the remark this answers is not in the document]
vastcharade: yeah, it helps to be more descriptive with the naming, even if it is more verbose.

src/utils/utils-contour.js
```diff
@@ -42,11 +45,14 @@ export const buildContourSQL = (state, mapBounds, isPolygons = false) => {
     flip_latitude = false,
     contour_value_field = "z",
     lat_field = "raster_lat",
-    lon_field = "raster_lon"
+    lon_field = "raster_lon",
+    interval
   } = data
 
+  const { intervalTypeFieldName = "is_major", minor } = interval
```
vastcharade, line 52: i know we have assurances from the immerse side that the major interval is divisible by minor, but I can't help but wonder if we should validate that still

example/exampleContour.html
```diff
@@ -68,56 +68,60 @@
...
+          lat_field: config.lat_field,
+          lon_field: config.lon_field,
+          agg_type: 'AVG',
+          fill_agg_type: 'AVG',
+          bin_dim_meters: config.bin_dim_meters,
+          neighborhood_fill_radius: 0,
+          fill_only_nulls: false,
+          flip_latitude: false,
+          contour_offset: 0,
+          interval: {
```
vastcharade, line 92: I personally like this better, tho it does bring into question whether some of the other properties should be grouped too.. i dunno. I think for typical uses defaults will be used, which would simplify this struct considerably.

example/exampleContour.html
```diff
@@ -68,56 +68,60 @@
...
+            intervalTypeFieldName,
```
vastcharade, line 93: `intervalTypeFieldName` is an ok name. Problem is that it still doesn't quite describe the output. The output is a boolean (or more simplistically, a bit flag) that determines whether or not it is a major or minor interval. Maybe it should be named "isMajorColumnName" or "isMajorField" or "isMajorTag" ... "intervalType" implies that the domain below should read:

```
  domain: ["major", "minor"]
  ...
```

As I write that down, I now wonder if that's what we should do -- use a string-type identifier, rather than an "is_major" true/false.

We can still do that in the `CASE` as the string will auto-convert to a string dictionary (as long as the table-function output can handle that -- experience thus far is it is a bit flaky yet, but in normal cases it would work just fine).

Something to consider. Otherwise looks good to me.

johallar: Makes sense about the `is*` naming, I can def do that bit

re: string identifiers, I feel like these identifiers would also have to be specified as part of `data`, or we'd be dealing with more implicit/magic strings ("major"/"minor"), which to me is starting to look a bit more complex than it needs to be/overly configurable.

johallar: I can do the string identifier way too (assuming it works with tf), just my 2 cents

vastcharade: When I write this down:

```
{
  type: "contour",
  intervals: {
    typeFieldName: "interval_type",
    major: <major interval>,
    minor: <minor interval>,
  }
}
...

color: {
  field: "interval_type",
  type: "nominal",
  scale: { 
    domain: ["minor", "major"], 
    range: ["#999999", "#999999"]
  }
}
```

makes some sense to me.

The `major`, `minor` keys in the `intervals` block could be used directly as field names to avoid further verbosity/redundancy, so if you could change those names to:

```
{
  type: "contour",
  intervals: {
    typeFieldName: "interval_type",
    intervalA: <major interval>,
    intervalB: <minor interval>,
  }
}
...

color: {
  field: "interval_type",
  type: "nominal",
  scale: { 
    domain: ["intervalA", "intervalB"], 
    range: ["#999999", "#999999"]
  }
}
```

if you wanted. It does give way to allowing more interval types, if one wants, tho under the limitation that they need to be divisible by a least-common-denominator.

Anyway, sorry, we're snowballing here in a way, but it's important to decide because such things are harder to change later.

johallar: Ahh using the keys as the values does make more sense, though still not immediately clear (to n00b me anyways) that that is whats happening. I like the flexibility angle though, i'll give this a shot

johallar: It may make life much easier if we eventually add user defined intervals

johallar: So actually (realized when implementing this) I think we would need to provide more information than just an interval number to be able to identify > 2 interval types or we'll run into collisions pretty quick, and i'm not sure i want to go down the road of how to resolve those/etc in this PR. 

I could see something in the future for custom intervals like:
{
  intervals: {
    typeFieldName: "interval_type",
    intervalA: [0, 50],
    intervalB: [50, 100]
    ....
  }
}```

johallar: Or if you had an idea of what the SQL would look like for having > 2 intervals: 
```
  intervals: {
    typeFieldName: "interval_type",
    intervalA: <major interval>,
    intervalB: <minor interval>,
    intervalC: <other interval>,
  }
```

I'm having a hard time figuring out what that would look like

src/utils/utils-contour.js
johallar: I'll add a little "validateContourState" utility to check for things

johallar: Added some validation of the state passed in for contours! Also didn't see anything obvious but is there another/ better way than just throwing a new Error() in charting? Looks like those would bubble up and get caught in immerse sagas/thunks and the chart would display the error, so should work.

example/exampleContour.html
johallar: Yeah i'm not sure if any others squarely fit here as its currently named, unless we move all of the contour TF params into here or rename again. These are easy to change down the line, but let me know if you have strong feelings about moving some into this block or renaming things.

example/exampleContour.html
johallar: @vastcharade I'd like to get this one merged ASAP, but you have requested changes still. If you can give it another pass through today that would be much appreciated! Given the timeline i'm going to defer the additional changes in the comments above until we have a need for them.

vastcharade: ok, deferment is a good plan. I'll revisit this now.

But just to clarify on one point, although something like this:

```
{
  intervals: {
    typeFieldName: "interval_type",
    intervalA: [0, 50],
    intervalB: [50, 100]
    ....
  }
}
```

would be ideal, we are stuck by the limitation that the contour function only takes a single contour interval as an argument. Not a set of intervals. So when dealing with multiple intervals, you have to take the smallest of the intervals and make sure the others are divisible by it.

Then you use the least-common-denominator interval for the contour interval argument, and then expand the `CASE` statement outside the table function to include the multiple variants, so if you had 4 defined intervals, the case would look like so:

```
CASE mod(cast(contour_values as int), <max interval>) WHEN 0 THEN "intervalA" WHEN 1 THEN "intervalB" WHEN 2 THEN "intervalC" ELSE "intervalD" END as interval_type
```

and then your scales would look like:

```
color: {
  field: "interval_type",
  type: "nominal",
  scale: { 
    domain: ["intervalA", "intervalB", "intervalC", "intervalD"], 
    range: [<color A>, <color B>, <color C>, <color D>]
  }
}
```

In any event, we can defer for now. I will look re-review what you've got.

johallar: 🙏 🙇

johallar: The big assumption would be that there is a common denominator to all of the interval sizes specified, which _I think_ is out the window when we're talking about custom interval ranges, but with only major/minor we are enforcing in immerse (and charting now)

src/utils/utils-contour.js
```diff
@@ -91,35 +105,52 @@ export const getContourMarks = (layerName, state) => [
         field: "y"
       },
       strokeColor: {
-        scale: "contour_color",
-        field: "is_major"
+        scale: CONTOUR_COLOR_SCALE,
+        field: state.encoding.color.field
```
vastcharade, line 109: so now that we've moved to defining the `field` explicitly in the vega lite json, we need to verify that "field" exists, otherwise we're prone to setting field incorrectly. If field doesn't exist in the color encoding, then it seems this would fill out to "undefined" (or crash if state.encoding.color doesn't exist), resulting in a backend render error. Similarly with stroke width below.

It seems that as of now we can assume that the field here for both must be the `isMajorFieldName`. Perhaps we can use that as a default for now too? Seems a little backwards to use that as a default, but it makes sense at the same time. Anyway, seems like we need some protection here. Otherwise looks good.

NOTE: this kind of stuff is why I had started developing a render-vega-lite module in the first place that would automatically handle logic for this, i.e. including the scales and validating them when requested rather than the dev having to account for it in their own individual layer.

vastcharade (approved):
Looks like there's some holes that need filling regarding validation yet, but I'm ok with landing this as is to get things unblocked and fill in the holes as a follow-up. Up to you @johallar

example/exampleContour.html
```diff
@@ -1,123 +1,127 @@
...
+          type: "poly",
+          strokeColor: "white",
+          strokeWidth: 0,
+          fillColor: "",
+          lineJoin: "miter",
+          miterLimit: 10
+        },
+        encoding: {
+          geocol: "contour_polygons",
+          color: {
```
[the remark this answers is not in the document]
johallar: Yeah I believe it uses `geocol` if provided for this, not totally clear on the magic behind this though

johallar:
Thanks for the thorough review, I think we could go pretty deep on validation, but for now if you pass an invalid spec its gonna break, and the structure of the spec passed from immerse is quite static so should minimize the risk here. 

I made https://heavyai.atlassian.net/browse/FE-16036 so I won't forgot to follow up once we have some breathing room

src/utils/utils-contour.js
johallar: Is that all the stuff you added with wind barbs? Maybe I can try out using it in the follow up ticket

vastcharade: yes, and it is going to be exposed through a more general utility with an upcoming PR (https://github.com/heavyai/heavyai-charting/tree/chris/render_vega_lite) with the goal of it probably turning into it's own repo. But I wouldn't advise trying to migrate over to it on your own quite yet unless you want to experiment. It is going to be a bit tricky, and delicate, because of so much existing code bleed with the previous format (as evidenced by the whole stacked-legend fiasco). I'm happy to give you some background on it when the dust settles if you want to assist with the migration.

example/exampleContour.html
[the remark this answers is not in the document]
johallar: ok my answer made no sense, i think this question is going over my head a bit... may need to trace it through the polygon mixin a bit to find out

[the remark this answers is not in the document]
vastcharade: yeah, seems like there's a hole here as well that will need filling. it can basically act the same as the line implementation tho. Again, I don't mind you landing as is and fixing in a follow-up.

How does this example color the polygons tho?

src/utils/utils-contour.js
johallar: Nice! I'd be happy to be a guinea pig at some point

example/exampleContour.html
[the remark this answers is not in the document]
johallar: AH, contours creates its own scales, and uses the `contour_value_field` as the field for polygons. It generates:
```
  "scales": [
    {
      "name": "contour_fill",
      "type": "linear",
      "domain": [0.0, 4000.0],
      "range": ["red", "blue"]
    }
  ],
...
"marks": [
    {
      "type": "polys",
      "from": {"data": "contourPolygonLayer"},
      "properties": {
        "x": {"field": "x"},
        "y": {"field": "y"},
        "fillColor": {"field": "contour_values", "scale": "contour_fill"},
        "strokeColor": "white",
        "strokeWidth": 0,
        "lineJoin": "bevel"
      },
      "transform": {"projection": "mercator_map_projection"}
    }
  ]
```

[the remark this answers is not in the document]
johallar: Gonna merge once tests pass, and we can continue convos in follow ups, i have one or two things left for contour charting

PR #596, opened 2022-11-17
[FE-16027] Use parser to create filters, and apply to contour query

johallar:
# Description
Since we can't use the parser to generate the sql for the contour TF, it also wasn't including the layer/global filters in the custom contour util. This takes _just_ the filters from the parser and adds those to the contour query

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/raster-layer-line-mixin.js
+++ b/src/mixins/raster-layer-line-mixin.js
@@ -340,7 +340,6 @@ export default function rasterLayerLineMixin(_layer) {
 
   _layer.__genVega = function({
     table,
-    mapBounds,
     filter,
     lastFilteredSize,
     globalFilter,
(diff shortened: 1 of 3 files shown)


src/mixins/raster-layer-line-mixin.js
```diff
@@ -364,7 +363,13 @@ export default function rasterLayerLineMixin(_layer) {
     let sql
     if (isContourType(state)) {
       validateContourState(state)
-      sql = buildContourSQL(state, mapBounds)
+      const filterTransforms = _layer
```
vastcharade, line 366: so, is this an attempt to replicate the `transform` component of the `parser.writeSQL` call below? It's hard for me to say because I'm unfamiliar with the `data-layer` module, but it seems on the surface that this is simplifies the transform parsing that the `parser` object does. Is that fair to say? I'm just wondering how closely this matches the transform parse and expression build in that `data-layer` module, and if it doesn't very closely match, I wonder if we could leverage just calling the `transform` related component here.

vastcharade (approved):
I spoke with @johallar offline and we agreed to approve this for now as a stop-gap to keep the contour iteration train moving.

We discussed an alternate plan that would be a more ideal way to implement this using the `data-layer` module with specialized table functions. On the surface it seems like what we discussed is a plausible idea, but approving this in order to keep iterating downstream.

johallar:
Ticket for posterity: https://heavyai.atlassian.net/browse/FE-16048

Plan is to at least use a custom parser for the sub query, and use the contour util to wrap it in the TF. Best case I can create a custom parser to generate the whole query including TF and sub-query

PR #599, opened 2022-11-21
Pulls the "render-vega-lite" components previously embedded in wind barbs into a general utility

vastcharade:
This is to enable general use, particularly in a follow up PR that lands the cross-section work.

The "render-vega-lite' components were first landed in this PR: https://github.com/heavyai/heavyai-charting/pull/585

The idea behind the render-vega-lite components is to have a more formal module that handles the parsing and vega generation from the json object passed to the raster layers from their respective `setState` methods.

What's currently in place seems to have evolved from a rather impromptu implementation, and does not have any formal schema/validation checking nor comments/documentation. 

The render-vega-lite module, on the other hand, is a first attempt at more closely aligning with [vega-lite](https://vega.github.io/vega-lite/docs/) in terms of schema/spec definitions. It is still in a loose form, but the hope is that we can eventually make it robust enough to match vega-lite in terms of documentation, parsing/validation, extensibility, and reliability. It's a decent start, and certainly feels like a more formal/central module to handle the parsing and vega generation.

Anyway, currently still only utilized by wind barbs, so this PR impacts nothing else, until the cross-section follow-up.

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


vastcharade:
@johallar @cmatzenbach 

Added you guys as reviewers just so I can land this. I do not expect you to do a formal review. All I ask is that the new file naming/organization mostly makes sense. It's really just pulling out code that was previously in `src/mixins/raster-layer-windbarb-mixin.js`

src/mixins/render-vega-lite/Definitions/Config/ConfigDefinitionInterface.js
```diff
@@ -0,0 +1,32 @@
+import PropertiesDefinitionInterface from "../PropertiesDefinitionInterface"
+
+export default class ConfigDefinitionInterface extends PropertiesDefinitionInterface {
+  /**
+   * @param {PropDescriptor} prop_descriptor
+   * @param {VegaPropertyOutputState} vega_property_output_state
+   */
+  materializeProperty(prop_descriptor, vega_property_output_state) {
```
johallar, line 8: As a general thing, javascript convention is to use camel case. Probably too much pain to convert everything you've done so far but for future things

vastcharade: sorry, to use camel case for what, the argument names, like `prop_descriptor`? 

Ok, does the FE have any formal coding style guides? I don't think I've ever seen any. This style is coming out of the BE side of things where class/struct/API methods are camel case, arguments and free functions are snake case. It's a form I'm used to and therefore prefer (with bias).

johallar: To your question, yes, all variable and function names are generally camel case.

Makes sense with your background, just pointing out that it is not the norm in javascript development. Variable and function names are camel case, class names upper camel case is the norm. Javascript itself uses this convention:
Screen Shot 2022-11-21 at 11 58 17 AM

There's no heavy style guide as far as I know, but definitely not a bad idea.

johallar: I guess that only shows function names, but take my word for it 😉

johallar: Also just to be clear, not asking for change, snake case def works and I've seen it before but I think we should use camel case going forward since I think this is more the norm + what we do in other heavy repos (mainly immerse, connector too)

@cmatzenbach any thoughts?

johallar (approved):

PR #606, opened 2022-11-29
[FE-16082] Check for geo points, adjust sql if they are sent

cmatzenbach:
Receives a new prop from the immerse pseudo-query that tells charting whether or not the dimensions are numerical or geo point types; adjusts sql query accordingly.

### DIFF
--- a/src/utils/utils-contour.js
+++ b/src/utils/utils-contour.js
@@ -52,7 +52,7 @@ export const buildContourSQL = ({
     contour_value_field = "z",
     lat_field = "raster_lat",
     lon_field = "raster_lon",
-    is_geo_point_type,
+    is_geo_point_type = false,
     intervals
   } = data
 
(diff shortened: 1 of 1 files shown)


johallar (approved):
LGTM, i'm not sure if we need to include `dist/charting.js` in this repo, but we are including it, so it should be updated with any changes: `npm run build`

src/utils/utils-contour.js
```diff
@@ -52,6 +52,7 @@ export const buildContourSQL = ({
     contour_value_field = "z",
     lat_field = "raster_lat",
     lon_field = "raster_lon",
+    is_geo_point_type,
```
johallar, line 55: maybe also default this value to false just to be explicit

johallar:
@cmatzenbach can you also commit the updated `dist/charting.js`?

PR #608, opened 2022-11-29
[FE-16091] Cross section chart does not render with parameterized selectors

cjvil:
Not ideal, but we need to accept strings as cross section coordinates, as parameterized values are only processed as values when query is run/in crossfilter. Immerse PR adds a check at that point instead.


src/mixins/render-vega-lite/Definitions/Transform/Transforms/CrossSection2dDefinitionObject.js
```diff
@@ -73,12 +73,16 @@ export default class CrossSection2dDefinitionObject extends PropertiesDefinition
...
+      (typeof obj.crossSectionLine[0][1] !== "number" &&
+        typeof obj.crossSectionLine[0][1] !== "string") ||
       !Array.isArray(obj.crossSectionLine[1]) ||
       obj.crossSectionLine[1].length !== 2 ||
-      typeof obj.crossSectionLine[1][0] !== "number" ||
-      typeof obj.crossSectionLine[1][1] !== "number"
+      (typeof obj.crossSectionLine[1][0] !== "number" &&
+        typeof obj.crossSectionLine[1][0] !== "string") ||
+      (typeof obj.crossSectionLine[1][1] !== "number" &&
+        typeof obj.crossSectionLine[1][1] !== "string")
```
johallar, line 85: This condition is so long it may be extra useful to make these shorter, not necessary but possible!

`!obj.crossSectionLine[0].every(p => ["number", "string"].includes(typeof p))`
`!obj.crossSectionLine[1].every(p => ["number", "string"].includes(typeof p))`

I think that comes out to the same... If everything is _not_ a number or a string it will throw the error

johallar (approved):

src/mixins/render-vega-lite/Definitions/Transform/Transforms/CrossSection2dDefinitionObject.js
johallar: golf:
`!obj.crossSectionLine.every(coord => coord.every(p => ["number", "string"].includes(typeof p)))

PR #609, opened 2022-11-30
Adds Cross Section-ready line draw tool to the suite of lasso tools

vastcharade:
Adds a new cross-section ready line draw/edit tool to the suite of existing lasso tool (circle, polyline, & lasso).

This cross section line tool is supported by a couple different classes:

1) `CrossSectionLineShapeHandler` class, embedded in `src/mixins/ui/lasso-tool-ui.js`
 a) this class supports building the cross-section line shape (currently limited to only a single-segment line) via various mouse click/move events
2) `LatLonPolyLineClass` in `src/mixins/ui/lasso-shapes/LatLonPolyLine.js`
 a) this class supports drawing the cross-section line in mercator-projected space, allowing to see the mercator-related bends/deformations when drawing a line with lat/lon coordinates.
 b) most of the functionality to do this was already in place. It just required pulling out some shared code from the `LatLonPoly` class to be used amongst both classes

When adding the lasso draw tool controls to a raster chart via the `addDrawControl()` method, a new argument can be passed to that method to control which draw utilities you want to see/use - circle, polyline, lasso, and/or cross section. This argument is bitflag controlled via the new `LassoToolSetTypes` bitflag enum class (`src/mixins/ui/lasso-tool-set-types.js`)

For example, to only see the new cross section tool in the draw tool contol ux, you would call:

```js
...
import * as HeavyCharting from "../src/index.js"
...
import LassoToolSetTypes from "../src/mixins/ui/lasso-tool-set-types.js"

const cross_section_chart = HeavyCharting.rasterChart(...)
...

cross_section_chart.addDrawControl(LassoToolSetTypes.kCrossSection)
```

Additional changes include a somewhat more refined event callback system allowing users to capture create/delete and edit start/end events, which are useful to capture when a cross-section line is created/deleted or modified (either moved, rotated, scaled, or individual vertices moved). That way you can easily update the the real cross section chart (or parameters) where/when appropriate.

Changes include a working example in `exampleLineCrossSection.html/js` (tho the server config is currently set to my local workstation)

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


johallar:
So cool to use this in the example!!

I noticed one thing, not sure if this is a prob with the example or with the draw tool itself:
1. Draw a line
2. Click the line tool again
3. You are able to draw 2 line segments

draw-tool

src/mixins/ui/lasso-shapes/LatLonViewIntersectUtils.js
```diff
@@ -36,6 +37,289 @@ function canIntersect(numerator, denominator) {
...
+      const denominator_t = denominator_t_functor()
+      if (denominator_t === 0) {
+        // colinear
+        // eslint-disable-next-line no-warning-comments
+        // TODO(croot): this needs filling out in the general case
+        // NOTE: this should never be hit in the LonLatPoly case because of
+        // the delta_x/delta_y === 0 check.
+
+        // eslint-disable-next-line no-console
+        console.assert(false, `Collinear intersection needs completing`)
```
johallar, line 198: Are you from the midwest? I've been told that "<x> needs <y>[ing | ed]" is a midwest thing

src/mixins/ui/lasso-shapes/LatLonViewIntersectUtils.js
```diff
@@ -36,6 +37,289 @@ function canIntersect(numerator, denominator) {
...
+          screen_pt = new_screen_pt
+        }
+        rtn_obj.subdivide = true
+
+        // insersection points must be stored by t in ascending order.
+        // There will be exactly 2 points in the end, so just need to check
+        // whether to insert this new intersection point at the front or back
+        // of the returned point array
+        if (t < current_t) {
+          rtn_obj.lonlat_pts.splice(0, 0, lonlat_pt)
```
johallar, line 265: You should be able to use `unshift` here instead of splice to add to the beginning of the array

src/mixins/ui/lasso-shapes/LatLonViewIntersectUtils.js
johallar: Oh copy paste from the midwest maybe...

johallar: WAIT A MINUTE
`TODO(croot): this needs filling out in the general case`

johallar (approved):
Code LGTM, one potential bug mentioned in other comment

vastcharade:
> 3\. You are able to draw 2 line segments

hmm, yeah, that needs to be fixed.

vastcharade:
> Tho need to decide on appropriate behavior here. Do we allow the ability to create a new line, and if they do it wipes out the currently existing one? Or do we prevent the user from creating a new line until any existing lines are deleted?

I've decided to go with the former. It makes more sense overall.

johallar:
I like option 1 as well

cmatzenbach (approved):
Same feedback as @johallar re: multiple line segments, but the only other thing I see is that backspace does delete a line, but the delete button does not.  Not sure if that's standard for our draw library or if that's something missing from this tool - I generally go for the delete button when doing things like this, but not a big deal if users are used to using backspace.

Other than that, LGTM.

vastcharade:
> the only other thing I see is that backspace does delete a line, but the delete button does not

Sorry, that was probably an oversight due to my testing it on a mac keyboard. I'll see about adding the "delete" key too.

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -733,14 +800,227 @@ class LassoShapeHandler extends ShapeHandler {
...
+          this.finishShape()
+          shapeBuilt = true
+        }
+      }
+
+      if (!shapeBuilt) {
+        this.enableBasemapDebounceFunc()
+        this.canvas.focus()
+        this.activeShape = null
+        // this.activeIdx = -1
```
cjvil, line 949: should this be removed?

vastcharade:
If y'all dont mind, I'm going to land this now and follow it up with an immediate PR that addresses the above bug. Makes more sense since the follow-up has a bunch of other code in it for completeness.

PR #611, opened 2022-12-02
Enforces cross-section line draw rule by only allowing one drawn cross section at a time

vastcharade:
Fixes the bug noted here: https://github.com/heavyai/heavyai-charting/pull/609#issuecomment-1334030150

Also includes an expansion of the event system to include some useful signals to help monitor situations where a cross-section line is deleted when a new cross section line is drawn. This was done as it may be useful to avoid re-drawing the cross section chart in that case and instead wait for the new line draw to complete. The client can check for that case with some new event extensions, namely the:

`LASSO_TOOL_TYPE_ACTIVATED`/`LASSO_TOOL_TYPE_DEACTIVATED` events
or the
`LASSO_TOOL_DRAW_STARTED`/`LASSO_TOOL_DRAW_ENDED` events

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/example/exampleLineCrossSection.js
+++ b/example/exampleLineCrossSection.js
@@ -233,18 +233,11 @@ function create_charts(
         LassoGlobalEventConstants.LASSO_SHAPE_CREATE,
         event_obj => {
           if (cross_section_layer) {
-            const { shape } = event_obj
-            if (shape instanceof LatLonPolyLine) {
-              // if the newly created lasso shape is a 'LatLonPolyLine', which is
-              // the current name of a cross section line shape, then attach an
-              // edit callback to it to capture any/all edits after creation.
-
-              // TODO(croot): find a better way to tag a shape as being a "cross-section line"
-              // Using 'instance of' may be too broad if we end up using LatLonPolyLine as a
-              // general tool for generating a line to filter results using a 'buffer'
-              // Also 'instance of' could be too narrow if we use another shape to draw a cross
-              // section other than LatLonPolyLine
-              // LatLonPolyLine will suffice for now as a differentiator.
+            const { shape, lasso_tool_type = null } = event_obj
+            if (lasso_tool_type === LassoToolSetTypes.kCrossSection) {
+              // if the newly created shape was created as a result of the
+              // CrossSection tool then attach an edit callback to it to
+              // capture any/all edits after creation.
 
               // NOTE: adding an edit event callback directly on a cross-section line shape is one
               // way to do this. An alternative way is to add a global edit event callback for all edited
--- a/example/exampleLineCrossSection.js
+++ b/example/exampleLineCrossSection.js
@@ -209,6 +209,25 @@ function create_charts(
         }
       }
 
+      let is_cross_section_enabled = false
+      pointMapChart.onDrawEvent(
+        LassoGlobalEventConstants.LASSO_TOOL_TYPE_ACTIVATED,
+        event_obj => {
+          if (event_obj.lasso_tool_type === LassoToolSetTypes.kCrossSection) {
+            is_cross_section_enabled = true
+          }
+        }
+      )
+
+      pointMapChart.onDrawEvent(
+        LassoGlobalEventConstants.LASSO_TOOL_TYPE_DEACTIVATED,
+        event_obj => {
+          if (event_obj.lasso_tool_type === LassoToolSetTypes.kCrossSection) {
+            is_cross_section_enabled = false
+          }
+        }
+      )
+
       // setup an event trigger for when a lasso shape is created
       pointMapChart.onDrawEvent(
         LassoGlobalEventConstants.LASSO_SHAPE_CREATE,
@@ -281,6 +300,15 @@ function create_charts(
             // For example, one approach is to set the line to be the bottom
             // edge of the current map view or the bottom edge of the raster
             // volume
+            // Tho you may want to ignore doing this if the shape was deleted
+            // as a result of a new line being drawn. You can check for this
+            // case by checking the "is_cross_section_enabled" variable that's
+            // tied to tool activate/deactivate signals
+            if (!is_cross_section_enabled) {
+              // no cross-section lines are currently drawn or being drawn on
+              // the map, so may want to reset the cross section endpoints to
+              // some default here.
+            }
 
             // turn off the croos-section line-local edit event callback
             // this is not really necessary, but it's good shutdown/cleanup
--- a/src/mixins/ui/lasso-event-constants.js
+++ b/src/mixins/ui/lasso-event-constants.js
@@ -30,13 +30,13 @@ export const LassoGlobalEventConstants = {
   /**
    * triggered when a new shape is being drawn via a lasso tool
    */
-  LASSO_TOOL_CREATE_STARTED: "lasso:tool:create:started",
+  LASSO_TOOL_DRAW_STARTED: "lasso:tool:draw:started",
 
   /**
    * triggered when a new shape draw ends for some reason,
    * either the shape is successfully completed, or cancelled
    */
-  LASSO_TOOL_CREATE_ENDED: "lasso:tool:create:ended",
+  LASSO_TOOL_DRAW_ENDED: "lasso:tool:draw:ended",
 
   /**
    * triggered when a new lasso shape is created
--- a/src/mixins/ui/lasso-event-constants.js
+++ b/src/mixins/ui/lasso-event-constants.js
@@ -28,6 +28,17 @@ export const LassoGlobalEventConstants = {
   LASSO_TOOL_TYPE_DEACTIVATED: "lasso:tool:deactivated",
 
   /**
+   * triggered when a new shape is being drawn via a lasso tool
+   */
+  LASSO_TOOL_CREATE_STARTED: "lasso:tool:create:started",
+
+  /**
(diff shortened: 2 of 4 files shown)


example/exampleLineCrossSection.js
```diff
@@ -209,23 +209,35 @@ function create_charts(
         }
       }
 
+      let is_cross_section_enabled = false
```
vastcharade, line 212: this is an example of how you can determine if a cross-section line that is deleted was deleted as a result of a new line draw.

example/exampleLineCrossSection.js
```diff
@@ -209,23 +209,35 @@ function create_charts(
...
-              // edit callback to it to capture any/all edits after creation.
-
-              // TODO(croot): find a better way to tag a shape as being a "cross-section line"
-              // Using 'instance of' may be too broad if we end up using LatLonPolyLine as a
-              // general tool for generating a line to filter results using a 'buffer'
-              // Also 'instance of' could be too narrow if we use another shape to draw a cross
-              // section other than LatLonPolyLine
-              // LatLonPolyLine will suffice for now as a differentiator.
+            const { shape, lasso_tool_type = null } = event_obj
+            if (lasso_tool_type === LassoToolSetTypes.kCrossSection) {
```
vastcharade, line 237: The new `LassoToolSetTypes` enum allows for a much more clean and flexible way to check if the shape event is associated with the cross-section line draw tool. In this case we pull the `lasso_tool_type` attribute, which will be available on most of the `LassoGlobalEventConstants`

src/mixins/ui/lasso-event-constants.js
```diff
@@ -13,6 +13,31 @@ export const LassoShapeEventConstants = {
 }
 
 export const LassoGlobalEventConstants = {
+  /**
+   * triggered when a new lasso tool is activated by pressing its
+   * corresponding button in the lasso tool set ui
+   */
+  LASSO_TOOL_TYPE_ACTIVATED: "lasso:tool:activated",
```
vastcharade, line 20: new signals. Still could use a way to document the properties of the object returned to callbacks.

src/mixins/ui/lasso-tool-set-types.js
```diff
@@ -24,53 +24,56 @@ export default class LassoToolSetTypes {
    * @type {Number}
    */
   // eslint-disable-next-line no-undef
-  static kCircle = new LassoToolSetTypes(1 << 0)
+  static kCircle = new LassoToolSetTypes(1 << 0, "circle")
```
vastcharade, line 27: Adding a descriptor string to the enum for easier prints.

src/mixins/ui/lasso-tool-set-types.js
```diff
@@ -24,53 +24,56 @@ export default class LassoToolSetTypes {
...
+    this.description = description
   }
 
   valueOf() {
     return this.value
   }
 
   toString() {
-    return `${this.value}`
+    return `${this.description !== null ? this.description : this.value}`
```
vastcharade, line 77: unfortunately `valueOf` trumps `toString` in cases where it's called like this:

```js
console.log(`${LassoToolSetTypes.kCrossSection}`)
```

so it seems you have to call the `toString()` method explicitly, which is no biggie IMO

```js
console.log(LassoToolSetTypes.kCrossSection.toString())
```

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -19,6 +19,92 @@ const MathExt = Draw.Math
...
 
+/**
+ * Enum class that defines the different contexts in which a shape handler's destroy() method
+ * might be called. The enum context can be used in ShapeHandler-derived detroy() method to
+ * destroy the appropriate shapes that may have been created by the handler. For example,
+ * temporary shapes may be created while creating a new shape (i.e. a simple circle for each
+ * vertex for visual emphasis). The shape handler may want to only delete the temporary shapes
+ * in its destroy() method and not its primary shape.
+ */
+class DestroyType {
```
vastcharade, line 30: Ended up going through the process of including a DestroyType enum class to identify the context in which `ShapeHandler::destroy()` methods are called. This is useful to include a reason for a `LASSO_TOOL_DRAW_ENDED` signal.

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -19,6 +19,92 @@ const MathExt = Draw.Math
...
+   */
+  // eslint-disable-next-line no-undef
+  static kReset = new DestroyType(4, "reset")
+
+  /**
+   * Returns true if the destroy type is a cancellable type (i.e. cancelled or reset)
+   * @param {*} destroy_type
+   * @return {Boolean}
+   */
+  static isCancellableDestroyType(destroy_type) {
```
vastcharade, line 81: utility used to classify whether the destroy type is ultimately a cancellable type, and therefore primary shapes need to be destroyed along with secondary shapes.

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -52,6 +138,23 @@ class ShapeHandler {
...
+    assert(
+      false,
+      `${ShapeHandler.name}::getLassoToolType() needs to be overridden by derived class`
+    )
+  }
+
+  fireEvent(event_signal, event_obj = {}) {
+    this.drawEngine.fire(event_signal, {
+      ...event_obj,
+      lasso_tool_type: this.getLassoToolType()
```
vastcharade, line 154: adds the `lasso_tool_type` property to the event object that can be read by callbacks

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -65,6 +168,10 @@ class ShapeHandler {
     shape.registerEvents(Object.values(LassoShapeEventConstants))
     this.drawEngine.addShape(shape, selectOpts)
     this.drawEngine.moveShapeToTop(shape)
+
+    // now tag the shape to the originating tool, which may be useful
+    // for callback handlers to know
+    shape.lasso_tool_type = this.getLassoToolType()
```
vastcharade, line 174: tags created shapes with the lasso tool type too, in case that's useful. again, would be great to document this somewhere .... but alas.

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -170,19 +287,52 @@ class ShapeHandler {
...
   }
+
+  /**
+   * Called to fire a tool-create-ended event signal, using a destroy type to provide
+   * some extra metadata as to why the draw stopped.
+   * @param {DestroyType} destroy_type Enum describing the context under why the draw stopped
+   */
+  notifyDrawStop(destroy_type) {
+    this.fireEvent(LassoGlobalEventConstants.LASSO_TOOL_DRAW_ENDED, {
+      ended_reason: destroy_type.toString()
```
vastcharade, line 317: and `ended_reason` property is added to this event signal object as a quick way to describe the reason for the draw end, in case that's useful.

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -246,23 +400,32 @@ class CircleShapeHandler extends ShapeHandler {
...
+   * Called under various contexts to destroy either temporary shapes that might have
+   * been created during the shape draw and optionally destroy the primary shape that
+   * may have been drawn
+   * @param {DestroyType} destroy_type Enum describing the context under which destroy()
+   *                                   was called.
+   * @returns {Boolean} returrns true if one or more shapes were deleted as part of the destroy
+   */
+  destroy(destroy_type) {
+    let is_shape_deleted = false
+    if (DestroyType.isCancellableDestroyType(destroy_type)) {
```
vastcharade, line 418: you only destroy primary shapes inside this `isCancellableDestroyType` check. All other derived `destroy` methods do something similar. Also returning whether shapes were destroyed or not. This indicates whether there was data to clear or not, indicating whether a draw-stopped signal should be thrown or not.

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -467,18 +659,10 @@ class PolylineShapeHandler extends ShapeHandler {
         )
       )
 
-      // NOTE: we're intentionally avoiding setting this.polyShape
```
vastcharade, line 470: workarounds like what this comment describes is no longer necessary since we can control whether primary shapes should be destroyed or not in the `destroy` method.

src/mixins/ui/lasso-tool-ui.js
```diff
@@ -1356,17 +1601,19 @@ export default class LassoButtonGroupController {
       )
     }
 
-    if (lassoToolSetTypes & LassoToolSetTypes.kCircle) {
-      add_handler("circle", CircleShapeHandler)
-    }
-    if (lassoToolSetTypes & LassoToolSetTypes.kPolyLine) {
-      add_handler("polyline", PolylineShapeHandler)
+    const available_lasso_tools = {
```
vastcharade, line 1604: a simpler way to add new tools in the future. Could allow for registering custom tools if that need ever arises.

johallar (approved):
Works great for me!

PR #612, opened 2022-12-05
[FE-16108] Add bbox filter for contour lines layer

johallar:
Polygon layer includes the bbox filter in `_genVega` and passes along, line layer does not so this moves the bbox query generation to a utility and uses in polygon and line raster layer mixins. 

This also fixes the query when using geo points instead of lat/lng, convo:
https://heavy-ai.slack.com/archives/C045DL4QM0D/p1670271981480729

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


src/mixins/raster-layer-line-mixin.js
```diff
@@ -366,6 +368,14 @@ export default function rasterLayerLineMixin(_layer) {
       const filterTransforms = _layer
         .getTransforms(table, filter, globalFilter, state, lastFilteredSize)
         .filter(f => f.type === "filter")
+      const bboxFilter = getContourBoundingBox(
+        state.data[0],
+        chart.map().getBounds()
```
cmatzenbach, line 373: This is the exact fn call that's getting messed up with custom dimensions.....sigh

src/utils/utils-contour.js
```diff
@@ -166,4 +166,20 @@ export const validateContourState = state => {
...
+    throw new Error("Latitude and Longitude fields must be provided in data array")
+  }
+}
+
+export const getContourBoundingBox = (data, mapBounds) => {
+  const table = data.table
+  const isGeoPoint = data.is_geo_point_type
+  const latField = isGeoPoint
+    ? `ST_Y(${table}.${data.lat_field})`
+    : `${table}.${data.lat_field}`
```
cmatzenbach, line 179: Doesn't _seem_ like it matters here, since you're just doing equality comparisons, but I've been wrapping the lat/lon fields in parens to safeguard against any applied calculations (since `lon - y` will break any calculation using division or multiplication due to order of operations).  I don't really see it being necessary here, but I'm wondering if it's safer just because I'm not sure of the various types of custom expressions users use.

src/utils/utils-contour.js
```diff
@@ -166,4 +166,20 @@ export const validateContourState = state => {
...
+  const lonField = isGeoPoint
+    ? `ST_X(${table}.${data.lon_field})`
```
cmatzenbach, line 181: same as above

cmatzenbach (changes requested):
Just safeguard those fields (if I find a better way to do it on the immerse side I'll take it out later) and good to go (pending local test)

src/utils/utils-contour.js
johallar: 👍 done

vastcharade (approved):
lgtm

PR #615, opened 2022-12-06
[FE-15866][FE-16067] integrate line select tool + prevent error on destroy cross section

cjvil:
Exports constants, guards against some conditions Immerse hits. Asserts for range removed as it is undefined when filterAll is called, which I doubt is new behavior. Added a noop for dispose as cross section seems to only have a dimension proxy.


israelvicars (approved):
Looks good 🚀

PR #619, opened 2022-12-09
[FE-16190] Fix SQL to accept custom measures

johallar:
SQL generation code was written with the assumption that the field passed in was a field name, but can be an expression. This changes to use a const for output field alias. 

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


vastcharade (approved):
lgtm

PR #621, opened 2022-12-14
[FE-16124] Contour popups match styles from layer

johallar:
Screen Shot 2022-12-14 at 12 12 49 PM
Screen Shot 2022-12-14 at 12 12 43 PM

Worlds ugliest chart?

popup-styles-contour

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge


cmatzenbach (approved):
LGTM

johallar:
@vastcharade if you're around today, would appreciate another set of eyes

src/utils/utils-vega.js
```diff
@@ -173,7 +175,11 @@ export function createVegaAttrMixin(
       rtnVal = layerObj[nullFunc]()
     } else if (input !== undefined && useScale) {
       const encodingAttrName = capAttrMap[capAttrName]
-      const capAttrObj = layerObj.getState().encoding[encodingAttrName]
+      let capAttrObj = layerObj.getState().encoding[encodingAttrName]
+      const scaleType = capAttrObj.type
```
vastcharade, line 179: seems like we should protect against the case that `capAttrObj` is `undefined` for the attr unless at this point we are assured that the `encoding` lookup by default returns `{}` for an attr.

src/utils/utils-vega.js
```diff
@@ -190,13 +196,23 @@ export function createVegaAttrMixin(
...
-        if (capAttrObj.type === "ordinal") {
+        if (scaleType === "ordinal") {
           ordScale.domain(domainVals).range(capAttrObj.range)
           // if color range is not in domain, it's an Other item
           rtnVal =
             domainVals.indexOf(input) === -1
               ? ordScale("Other")
               : ordScale(input)
+        } else if (scaleType === "nominal") {
+          let formattedInput = input
```
vastcharade, line 207: can you describe the use case here as it pertains to needing to convert an input to a boolean? I understand that if all the domains for the scale are boolean, then the input value needs to be boolean, but it feels like you're protecting against a special use case here. I guess what I'm asking is, what's the use case you've observed where the `input` is not a boolean, and therefore needs to be converted, when the `domain` values are boolean (implying at most a true and false domain)?

src/utils/utils-vega.js
```diff
@@ -190,13 +196,23 @@ export function createVegaAttrMixin(
...
+          if (domainVals.every(v => typeof v === "boolean")) {
+            formattedInput = Boolean(input)
+          }
+          const indexOfVal = domainVals.indexOf(formattedInput)
```
vastcharade, line 211: you are essentially conducting your own `ordinal` scale lookup here. Any reason you're doing this manually rather than using the d3 ordinal scale?

vastcharade (approved):
logically looks fine. i think you need to protect against the case where an attr encoding doesn't exist (unless that protection is guaranteed that I'm not seeing).

Also had some clarifying questions.

I'm approving preemptively so that you can land when appropriate to do so.

src/utils/utils-vega.js
johallar: Nope good call, maybe this can even just fall into the same block as ordinal

src/utils/utils-vega.js
johallar: The special case _was_ when the vega lite like spec was passing true/false for the domain, but the actual values returned from the query were 0/1... not sure if this is the case anymore, digging in a bit

src/utils/utils-vega.js
vastcharade: yeah, it probably can (other than the boolean type conversion -- tho it's possible that could just work)

johallar: Agh, yeah it needs this conversion to handle getting 0/1 as an input when the domain is boolean, but otherwise I can roll it in with the previous condition

johallar: I take it back, 0/1 does work with a d3 ordinal scale of false/true! I'm going to keep this as a separate case though, because it doesn't handle "Other" the same as the "ordinal" vega scale type

johallar: I double take it back, this works in newer versions of d3 but not ours, i'm going to keep boolean logic here as well

johallar: Ok, final state is that "nominal" scale is its own block, it does the manual boolean conversion if the domain is all booleans, but it uses the d3 ordinal scale

PR #626, opened 2023-01-09
[FE-16237] Set new prop to avoid filtering on clicked poly for contour

cmatzenbach:
Previously polygon layer had a default `onClick` function which filters by the poly selected.  This is not wanted for contour (especially since it interferes with the cross-section line tool), so adding a new settable property to the layer which can be used to bypass this function.

This fixes an issue with the line tool in contour where the `onClick` handler for this layer would fire when a user clicks to start or end a line, and filter out all the results from the data set.


johallar (approved):
🎉 Looks great!

PR #627, opened 2023-01-09
[FE-16290] Contour multilayer support

johallar:
In order to have multiple contour charts in one layer, we need to name the scales on a per layer basis. This includes the layer name for each scale so that we can have multiple color, width scales in the same vega spec. 

This also refactors a small bit of legend code to make sure the domain is an array, and make the check a little simpler


cmatzenbach (approved):
LGTM!

PR #630, opened 2023-01-25
[FE-16355] Safety optional chaining, validate without throwing error

johallar:
# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/chart-addons/stacked-legend.js
+++ b/src/chart-addons/stacked-legend.js
@@ -2,6 +2,7 @@ import * as _ from "lodash"
 import { getRealLayers } from "../utils/utils-vega"
 import assert from "assert"
 import { format } from "d3-format"
+import { logger } from "../utils/logger"
 
 const hasLegendOpenProp = color =>
   typeof color.legend === "object" && color.legend.hasOwnProperty("open")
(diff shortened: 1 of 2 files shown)


johallar:
@vastcharade A bandaid for now, the legend still renders correctly and we don't throw an error causing the whole chart to fail. In this case the next render (not ideal...) has the scales correctly populated from vega metadata.

src/chart-addons/stacked-legend.js
```diff
@@ -504,6 +507,24 @@ function legendState_v2(state, useMap) {
...
+  if (!Array.isArray(state.domain)) {
+    errors.push(`Domain must be an Array, but was ${JSON.stringify(state.domain)}`)
+  }
+  if (state.domain.length !== 2) {
+    errors.push(
+      `Domain must have length of exactly 2, but was ${state.domain.length}`
+    )
+  }
+  if (typeof state.domain[0] !== "number") {
+    errors.push(`Domain must be array of integers but was ${JSON.stringify(state.domain)}`)
```
vastcharade, line 524: can be any number - not just integers

vastcharade (approved):
lgtm -- curious, any progress in determining the root cause?

johallar:
> lgtm -- curious, any progress in determining the root cause?

A couple things from investigation yesterday:
1. There are multiple `_doRender` calls for any multilayer chart, so its not specific to these repro steps
2. One reproduction of the issue happens when you select an individual layer, and click back to the master layer before the request has completed. Seems like the single layer request completes, while the master layer is loading, which only has 1 of the 2 scales needed. And then when it tries to render the legend for both layers, it runs into this error.

PR #639, opened 2023-05-11
[FE-16789] Multilayer cross section setup

cmatzenbach:
New mixin, vega-lite setup, all the charting plumbing needed to render 1-D cross section


src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+  )
+
+  _layer.__genVega = function({
+    chart,
+    table,
+    filter,
+    lastFilteredSize,
+    globalFilter,
+    // eslint-disable-next-line no-unused-vars
+    pixelRatio,
```
johallar, line 127: Can probably remove

src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+import CrossSectionTerrainConfigDefinitionObject from "./render-vega-lite/Definitions/Mark/CrossSectionTerrainConfigDefinitionObject"
+import { materializePropDescriptors } from "./render-vega-lite/RenderVegaLite"
+
+// only used for jsdoc
+// eslint-disable-next-line no-unused-vars
+import VegaPropertyOutputState from "./render-vega-lite/VegaPropertyOutputState"
+
+// eslint-disable-next-line no-warning-comments
+// TODO(croot): this seems like it is used in at least one other layer mixin. Make into a utility somewhere
+function create_post_filter_transform(post_filters) {
```
johallar, line 27: Camel case if poss

src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+    return state
+  }
+
+  // NOTE: as of 11/14/22 the "getTransforms" method that is found in most of the other raster layer mixin classes
+  // only seems to be called from immerse via the buildRasterExportSql() method in raster-sql.ts
+  //
+  // It is not currently believed that Mesh2d marks need to support export, so the "getTransforms" method is not
+  // going to be included
+
+  const prop_descriptors = new Map()
```
johallar, line 72: Samesies (camel case), if you have it in you to rename things throughout the new mixin 😅

src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+    )
+    const {
+      sql_parser_transforms,
+      vega_data_formats,
+      vega_transforms,
+      vega_scales,
+      mark_properties
+    } = _vega_property_output_state.flatten()
+
+    assert(vega_data_formats.length <= 1)
```
johallar, line 152: Good opportunity to add a message if this assertion fails, I _think_ the API is just `assert(vega_data_formats.length <= 1, "This is why it failed")`

src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+      sql_parser_transforms.push({
+        type: "filter",
+        expr: globalFilter
+      })
+    }
+
+    const data = [
+      {
+        name: layerName,
+        ...(vega_data_formats.length ? vega_data_formats[0] : {}),
```
johallar, line 179: minor, but can make this a little simpler: `vega_data_formats.?[0] ?? {}`

src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+      ...vega_transforms
+    ]
+
+    const marks = [
+      {
+        type: CrossSectionTerrainConfigDefinitionObject.key,
+        from: {
+          data: layerName
+        },
+        properties: Object.assign({}, ...mark_properties)
```
johallar, line 196:
```suggestion
        properties: { ...mark_properties}
```

src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+    return createCrossfilterDimensionProxy(new_val)
+  })
+
+  let _vega = null
+  const _cf = null
+  _layer.crossfilter = createRasterLayerGetterSetter(_layer, _cf)
+
+  // eslint-disable-next-line no-unused-vars
+  _layer._genVega = function(chart, layerName, group, query) {
+    const realLayerName = layerName
```
johallar, line 247: Can remove

src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+      return new_val
+    }
+    return createCrossfilterDimensionProxy(new_val)
+  })
+
+  let _vega = null
+  const _cf = null
+  _layer.crossfilter = createRasterLayerGetterSetter(_layer, _cf)
+
+  // eslint-disable-next-line no-unused-vars
```
johallar, line 245: Remove unused vars + lint disable line

src/mixins/raster-layer-cross-section-terrain-mixin.js
```diff
@@ -0,0 +1,325 @@
...
+      table: _layer.crossfilter().getTable()[0],
+      filter: _layer.crossfilter().getFilterString(realLayerName),
+      globalFilter: _layer.crossfilter().getGlobalFilterString(),
+      lastFilteredSize: lastFilteredSize(_layer.crossfilter().getId()),
+      pixelRatio: chart._getPixelRatio()
+    })
+    return _vega
+  }
+
+  // _layer.getPrimaryColorScaleAndLegend = function() {
```
johallar, line 260: Remove if unused

src/mixins/render-vega-lite/Definitions/Mark/CrossSectionTerrainConfigDefinitionObject.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * @param {Object} definition_object
+   * @param {ParentInfo} parent_info
+   */
+  constructor(definition_object, parent_info) {
+    super(definition_object, parent_info)
+
+    this.strokeColor_ =
```
johallar, line 23: If you didn't want to, we don't have to have all of this shenanigans to try and make these "private". It could just set the property on `this` and remove the `get` stuff below: `this.strokeColor = definition_object.strokeColor`

src/mixins/render-vega-lite/Definitions/Transform/Transforms/CrossSectionTerrainDefinitionObject.js
```diff
@@ -0,0 +1,320 @@
...
+  // eslint-disable-next-line no-undef
+  static key = "cross_section1d"
+
+  /**
+   * @param {Object} definition_object
+   * @param {ParentInfo} parent_info
+   */
+  constructor(definition_object, parent_info) {
+    super(definition_object, parent_info)
+    assert(
```
johallar, line 20: Opportunity for more meaningful error message

src/mixins/render-vega-lite/Definitions/Transform/Transforms/CrossSectionTerrainDefinitionObject.js
```diff
@@ -0,0 +1,320 @@
...
+      )
+    }
+    this.z_ = obj.z
+
+    /**
+     * @type {Number[]}
+     */
+    this.cross_section_line_ = []
+    if (
+      !Array.isArray(obj.crossSectionLine) ||
```
johallar, line 71: geez

johallar, line 71: Sheesh, could be a good opportunity to make this more readable and make a `validateCrossSectionLine` method or something. Not required

```
!Array.isArray(obj.crossSectionLine[1]) ||
      obj.crossSectionLine[1].length !== 2 ||
      !obj.crossSectionLine[1].every(p =>
        ["number", "string"].includes(typeof p)
      )```  
could def be pulled out into a function and called for each set of coords in the `crossSectionLine`

File: src/mixins/raster-layer-cross-section-terrain-mixin.js (hunk 3 above)
Comment (cmatzenbach) [reply src/mixins/raster-layer-cross-section-terrain-mixin.js:72]: I'm fine with function names (like `create_post_filter_transforms`), but every variable throughout these new mixins and his sql library is written in c++ style, I don't have the energy to change them all 😆 
The mix of both styles is bothering me though, so may go back through and refactor when I have the time

File: src/mixins/raster-layer-cross-section-terrain-mixin.js (hunk 4 above)
Comment (cmatzenbach) [reply src/mixins/raster-layer-cross-section-terrain-mixin.js:152]: To be honest, I don't even understand what this check is for, I'd have to spend a lot more time digging through his vega-lite building library to figure out how all these different components work.  It's not clear without a lot of debugging due to lack of comments.

File: src/mixins/raster-layer-cross-section-terrain-mixin.js (hunk 5 above)
Comment (cmatzenbach) [reply src/mixins/raster-layer-cross-section-terrain-mixin.js:179]: Good call!

File: src/mixins/raster-layer-cross-section-terrain-mixin.js (hunk 7 above)
Comment (cmatzenbach) [reply src/mixins/raster-layer-cross-section-terrain-mixin.js:247]: I'm unsure about this one.  What he did is very explicit, and makes me wonder if some modification happens to `layerName` when it is passed into the `__genVega` call that mutates the referenced var, thereby messing it up if you pass the same variable to `getFilterString()`.  I don't think it's worth it.

Comment (johallar) [reply src/mixins/raster-layer-cross-section-terrain-mixin.js:247]: Layer name is a string primitive so references won't be a thing here it will just receive the string value
Screen Shot 2023-05-12 at 10 23 00 AM

and changing a parameter reference won't change the  reference of the variable passed in. 
Screen Shot 2023-05-12 at 10 33 46 AM

If we're imagining that `layerName` is an object and properties of it may be mutated, then this re-assignment of the reference does nothing to prevent that.
Screen Shot 2023-05-12 at 10 35 49 AM

But if you want to choose fear over love than I understand 😉

File: src/mixins/raster-layer-cross-section-terrain-mixin.js (hunk 6 above)
Comment (cmatzenbach) [reply src/mixins/raster-layer-cross-section-terrain-mixin.js:188]: So changing this is actually what caused the bug I mentioned, where the `mark` object was transformed into an array-like-object (basically all nested objects were proceeded with a stringified numeric index).  I'm honestly not too sure why, chatGPT gave me this highlighting the differences:

When using the object spread syntax ({ ...mark_properties}), it does not handle nested objects correctly. Instead of creating deep copies of the nested objects, it only performs a shallow copy, which leads to unexpected behavior when modifying the nested objects. This behavior is inherent to the object spread syntax and is known as shallow copying.

On the other hand, Object.assign({}, ...mark_properties) is more suitable for copying objects with nested structures. It performs a shallow copy of the top-level properties, but it recursively performs a shallow copy of the nested objects as well, ensuring that the resulting object maintains the correct structure.

```
const mark_properties = {
  prop1: 'value1',
  nestedObj: {
    prop2: 'value2',
  },
};

const spreadCopy = { ...mark_properties };
const assignCopy = Object.assign({}, mark_properties);

mark_properties.prop1 = 'new value';
mark_properties.nestedObj.prop2 = 'new value';

console.log(spreadCopy); // { prop1: 'value1', nestedObj: { prop2: 'new value' } }
console.log(assignCopy); // { prop1: 'value1', nestedObj: { prop2: 'value2' } }
```
I don't think this example really highlights what's going wrong here, but does give good indication that they are not equivalent expressions in javascript.  Going to leave it as Chris R originally had it.

src/mixins/raster-layer-cross-section-terrain-mixin.js
cmatzenbach: While I completely agree with you, it's just seems really strange that he would explicitly make that variable if there was no need.....honestly I don't see what could be wrong with just using the same variable (everything you said makes perfect sense), but for now at least, I'm sadly going to choose fear in lieu of the tiny chance of introducing some hard-to-detect bug in the future :)

src/mixins/raster-layer-cross-section-terrain-mixin.js
johallar: Sneaky! I glossed over the spread operator there, makes sense this is spreading all the objects within that object.

israelvicars (approved):
I've done a cursory review after discussing implementation with @cmatzenbach . I think we should merge sooner so we can begin testing earlier. I know there are unresolved comments remaining. Maybe we could create a follow up task for additional review.

PR #641, opened 2023-05-23
[FE-16795] [FE-16796] Add second y-axis, ensure y-axes alignment

cmatzenbach:

Adds support for adding an additional y-axis to the coordinate-grid-raster-mixin, currently used by backend scatter, cross section, and cross section terrain.  Modifies the mixin and its HTML generation process to support right y-axis rendering when `useTwoYAxes` is set.  Additionally, adds a `chart.y2` prop, which is then set in the presence of a multi-layered cross section chart based on the terrain layer.  While this currently will always ensure the terrain layer gets the y2 axis, this should allow any layer to take the y2-axis with slight future modifications.  

Closes [FE-16795](https://heavyai.atlassian.net/browse/FE-16795) [FE-16796](https://heavyai.atlassian.net/browse/FE-16796)

### :mag: Verification Steps

1. Create a cross section chart, make sure only left y-axis appears with correct scale
2. Add a terrain layer, make sure only left y-axis appears with correct scale for layer
3. Switch to master tab, ensure multi-layered chart is rendered and a new y-axis appears on the right
4. Ensure left y-axis has the correct scale for normal CS chart, while right y-axis has the correct scale for terrain chart

### :camera_flash: Screenshot
2023-05-25-122808_3418x1915_scrot

[FE-16795]: https://heavyai.atlassian.net/browse/FE-16795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FE-16796]: https://heavyai.atlassian.net/browse/FE-16796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scss/chart.scss
+++ b/scss/chart.scss
@@ -457,15 +457,6 @@ body {
         }
     }
 
-    /*div.y2-axis-label, div.axis-label-edit.type-y2 {
-        transform-origin: 50% 0;
-        transform:translateX(-50%) rotate(90deg) ;
-        input {
-            cursor: vertical-text;
-            top: -1px;
-        }
-    }*/
-
 
     div.x-axis-label, div.axis-label-edit.type-x {
         transform: translateX(-50%);
--- a/scss/chart.scss
+++ b/scss/chart.scss
@@ -457,6 +457,15 @@ body {
         }
     }
 
+    /*div.y2-axis-label, div.axis-label-edit.type-y2 {
+        transform-origin: 50% 0;
+        transform:translateX(-50%) rotate(90deg) ;
+        input {
+            cursor: vertical-text;
+            top: -1px;
+        }
+    }*/
+
 
     div.x-axis-label, div.axis-label-edit.type-x {
         transform: translateX(-50%);
--- a/scss/chart.scss
+++ b/scss/chart.scss
@@ -437,7 +437,7 @@ body {
         fill: $gray5 !important;
     }
 
-    div.x-axis-label, div.y-axis-label, div.axis-label-edit {
+    div.x-axis-label, div.y-axis-label, div.y2-axis-label, div.axis-label-edit {
         position: absolute;
         font-size: 13px !important;
         color: $text-gray;
@@ -447,7 +447,7 @@ body {
         max-width: 100%;
     }
 
-    div.y-axis-label, div.axis-label-edit.type-y {
+    div.y-axis-label, div.axis-label-edit.type-y, div.y2-axis-label, div.axis-label-edit.type-y2 {
         transform-origin: 50% 0;
         transform:translateX(-50%) rotate(270deg) ;
         left: 4px;
@@ -2305,6 +2305,13 @@ body {
     }
 
     &.type-y .axis-input {
+        transform: translate(calc(-100% - 8px), -50%);
+        input {
+            text-align: right;
+        }
+    }
+
+    &.type-y2 .axis-input {
         transform: translate(calc(-100% - 8px), -50%);
         input {
             text-align: right;
--- a/src/charts/raster-chart.js
+++ b/src/charts/raster-chart.js
@@ -294,7 +294,7 @@ export default function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
   }
 
   _chart.getAllLayerTypes = function() {
-    return _layers.map(layer => layer?.layerType())
+    return _chart.getAllLayers().map(l => l?.layerType())
   }
 
   _chart.xRangeFilter = function(filter) {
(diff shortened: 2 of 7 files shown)


scss/chart.scss
```diff
@@ -457,6 +457,15 @@ body {
         }
     }
 
+    /*div.y2-axis-label, div.axis-label-edit.type-y2 {
```
johallar, line 460: Remove commented out stuff if not needed

src/charts/raster-chart.js
```diff
@@ -513,6 +544,106 @@ export default function rasterChart(parent, useMap, chartGroup, _mapboxgl) {
...
+              range = range.flatMap(r => r)
+            }
+            layer.layerType() === "crossSectionTerrain"
+              ? y2Ranges.push(range)
+              : yRanges.push(range)
+          }
+        }
+      }
+
+      if (xRanges.length) {
```
johallar, line 581: If this is a direct copy/pasta from `_updateXAndYScales` it would probably be better to move this logic to a function so that changes don't need to be made in two places down the line.

src/mixins/coordinate-grid-raster-mixin.js
```diff
@@ -1023,20 +1050,40 @@ export default function coordinateGridRasterMixin (_chart, _mapboxgl, browser) {
...
       yLabel = root.append("div")
-        .attr("class", "y-axis-label")
+        .attr("class", `${axisClass}-axis-label`)
     }
 
     if (text !== "") {
       // TODO(croot): should add the rotation and labelXPosition here
       // As of now (09/02/2016) the chart.css is breaking this.
 
+      if (axisClass === "y2") {
```
johallar, line 1083: Not necessary but consts for these axis keys would be a good way to go IMO, as they're used/checked a few times throughout

src/mixins/label-mixin.js
```diff
@@ -121,8 +121,15 @@ export default function labelMixin(chart) {
       chart.legend().legendType() === "quantitative"
 
     const iconPosition = {
-      left: type === "y" ? "" : `${getXaxisLeftPosition(hasLegend)}px`,
-      top: type === "y" ? `${getYaxisTopPosition()}px` : ""
+      left:
+        type === "y" || type === "y2"
```
johallar, line 125: style: could use `includes` syntax here

src/mixins/scatter-mixin.js
```diff
@@ -57,20 +60,32 @@ export default function scatterMixin(_chart, _mapboxgl, mixinDraw = true) {
...
           addDimAndRange(layer.xDim(), xDims, xRanges)
         }
 
-        if (typeof layer.yDim === "function") {
-          addDimAndRange(layer.yDim(), yDims, yRanges)
+        if (
+          _chart.useTwoYAxes() &&
+          layer.layerType() === "crossSectionTerrain"
+        ) {
+          if (typeof layer.yDim === "function") {
```
johallar, line 82: Could this `typeof layer.yDim === "function"` condition just be rolled up into the top `if`? It's checked separately in both the `if` and `else` it seems

src/mixins/scatter-mixin.js
```diff
@@ -101,6 +116,7 @@ export default function scatterMixin(_chart, _mapboxgl, mixinDraw = true) {
   }
 
   _chart.getDataRenderBounds = function() {
+    // TODO(C): THIS COULD ALSO BE WHERE THIS IS HAPPENING
```
johallar, line 119: old comment?

johallar (approved):
Awesome, working as expected with the immerse branch, dealers choice for suggested changes. This one probably top for me: https://github.com/heavyai/heavyai-charting/pull/641/files#r1205859671

src/charts/raster-chart.js
cmatzenbach: I actually am debating if the second part (`else` block) is used at all, and if it needs to stay.  Looking at the code, it really shouldn't even run - there are several problems:

- `getLayers()` is not declared as a stand-alone function anywhere in this file - it should be `_chart.getLayers()`
- `for (layer in layers)` isn't valid js - `layer` needs to be declared
- `viewBoxDim` isn't declared anywhere but is assigned a value

Maybe some of that stuff would pass in non-strict mode or whatever, but I don't see how the `getLayers()` call would work.  I was debating pulling it out entirely, but you know me and my extreme risk-adverse nature....

src/mixins/scatter-mixin.js
cmatzenbach: Definitely, good call

johallar: Hm... now i'm second guessing this, if it's in the top level `if` then if it falls into the `else` it will mean it's not a function. Seems like this could be condensed somehow tho haha maybe an `if` wrapping the parent `if`... not necessary

PR #645, opened 2023-05-31
[FE-16811] Add opacity prop descriptor for terrain mixin

cmatzenbach:
Add needed prop descriptor to render opacity in vega json for terrain layer


johallar (approved):
lgtm!

PR #646, opened 2023-06-02
[FE-16826] Add config options for d3 scale type

cmatzenbach:
Adds new API to allow log scales to be set for charts' y-axes


johallar (approved):
Same here w the scale constants, but too much to change right now, LGTM!

PR #649, opened 2023-06-06
[FE-16842] Show legend in bottom left for mesh2d layers (cross section)

johallar:
😌  The position was all driven from the `useMap` argument passed to raster chart. This enabled individual layers to determine their own legend position but falls back to the `useMap` logic

This also calls the same format function that legend v1 calls so we don't get crazy precision in the legend values (eg. 45.123853434534)

Before
Screen Shot 2023-06-06 at 5 07 32 PM

After:

Legend Position:
Screen Shot 2023-06-06 at 5 01 05 PM

Format legend values:
Screen Shot 2023-06-06 at 5 06 48 PM

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/chart-addons/stacked-legend.js
+++ b/src/chart-addons/stacked-legend.js
@@ -5,8 +5,8 @@ import { format } from "d3-format"
 import { logger } from "../utils/logger"
 
 export const LEGEND_POSITIONS = {
-  TOP_RIGHT: 'top-right',
-  BOTTOM_LEFT: 'bottom-left'
+  TOP_RIGHT: "top-right",
+  BOTTOM_LEFT: "bottom-left"
 }
 
 const hasLegendOpenProp = color =>
@@ -117,7 +117,7 @@ export function getLegendStateFromChart(chart, useMap, selectedLayer) {
       .filter((d, i) => i !== selectedLayer.currentLayer)
       .remove()
   }
-  
+
   return toLegendState(
     getRealLayers(chart.getLayerNames()).map(layer_name => {
       const layer = chart.getLayer(layer_name)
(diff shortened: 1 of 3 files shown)


cmatzenbach (changes requested):
So legend position looks good now, and isn't displaying crazy long decimals, but I'm not sure the values it's displaying now make a ton of sense.  See the attached screencast - it starts with "440m" (not sure what the "m" is referring to - meters?), goes up at seemingly random intervals, and in the middle has a 0 overlayed that prevents you from seeing the values in a few boxes.


https://github.com/heavyai/heavyai-charting/assets/15268289/fa0ce3e2-40c3-4a27-a3f7-e4211a7edaa9

cmatzenbach (approved):

johallar:
Hm, yeah we've never had a stacked legend on a coordinate grid chart which is probably why we haven't seen the `0` hover thing before. That's coming from the axes, where you can manually input min/max values.

As far as the `m`, this uses the same formatting function as the rest of the legends so I think this is how immerse just currently formats numeric legends. Scatter chart on master:
Screen Shot 2023-06-08 at 10 44 24 AM
Screen Shot 2023-06-08 at 10 47 24 AM

Wondering if the `m` is coming from "milli" prefix:
Screen Shot 2023-06-08 at 10 46 53 AM

I'll give both of these a quick look, but I think re-positioning and formatting still mostly a win with some good call outs for weird behavior interactions that we should look into separately.

johallar:
Put this bug in to follow up on the 0 overlap piece. Since the rest of our legends use the same format as the `m` bit i'm calling that "working as expected". Larger app considerations for changing how we format those values across the app for that one.

PR #653, opened 2023-08-30
[FE-17159] Changing the style of the map is affecting layer ordering

cjvil:
Positioning layer wasn't passed when re-adding render layer on style change. This should place the render layer under the label layer.


johallar:
My bad, but we shouldn't bump in immerse until https://github.com/heavyai/immerse/pull/8941 gets in. There is a a breaking change in https://github.com/heavyai/heavyai-charting/pull/652, until that immerse PR gets in.

johallar (approved):

PR #654, opened 2023-09-11
[FE-17163] Add forceResize property

cmatzenbach:
Adds forceResize property, which will bypass width/height delta checks to force a resize (needed for layer drawer, as chart will have same w/h but needs a resize since mapbox is detecting the size of the actual canvas element and will thus leave a gray area once the drawer is gone).

Now also includes hack to get mapbox to cooperate when we have an overlay drawer rendered overtop of it.

### DIFF
--- a/src/mixins/map-mixin.js
+++ b/src/mixins/map-mixin.js
@@ -73,6 +73,7 @@ export default function mapMixin(
   let _interactionsEnabled = true
   let _shouldRedrawAll = false
   let _forceResize = false
+  let _overlayDrawerOpen = false
 
   _chart.useLonLat = function(useLonLat) {
     if (!arguments.length) {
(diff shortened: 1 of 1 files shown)


src/mixins/map-mixin.js
```diff
@@ -771,7 +787,30 @@ export default function mapMixin(
...
       _lastHeight = height
-      _map.resize()
+
+      // this is a dumb hack, but it works and there doesn't seem to be another sensible way.
+      // problem is, mapbox looks to the size of "mapboxgl-canvas-container" to determine the
+      // render size of the canvas, regardless of what we feed it above. if there is an overlay
+      // drawer open, even if it sits at a higher z-index, the canvas size will be calculated
+      // according to the physical space in the DOM. thus, force the canvas container to be the
+      // size of the chart (which ignores the overlay), and reset it once mapbox is done resizing
+      if (_overlayDrawerOpen) {
```
johallar, line 797: Not gonna say I'm into it...but do what you have to do I suppose. Maybe clarify what you mean by "physical space in the DOM" in your comment, does that mean visible size or the whole element's size visible or not?

johallar (approved):

PR #655, opened 2023-10-09
[FE-17191] Join to geo support

johallar:
Big Changes:
* Updates transforms for line and poly mixins, so that if the datasource is a join it will group by the geometry table's `rowid`
* Line map wasn't using the join datasource properly, if it detects that the layer uses a join datasource it uses `table`which is the join datasource parameter, which will be processed to the full syntax down the line. 
* Use full `table.column` expression for rowid because for join data sources both tables have a rowid. 
* Some syntax improvements

# Merge Checklist
## :wrench: Issue(s) fixed:
- [ ] Fixes #0

## :smoking: Smoke Test
- [ ] Works in chrome
- [ ] Works in firefox
- [ ] Works in safari
- [ ] Works in ie edge
- [ ] Works in ie 11

## :ship: Merge

### DIFF
--- a/src/mixins/raster-layer-line-mixin.js
+++ b/src/mixins/raster-layer-line-mixin.js
@@ -1,3 +1,4 @@
+/* eslint-disable complexity */
 import {
   createRasterLayerGetterSetter,
   createVegaAttrMixin,
--- a/src/mixins/raster-layer-line-mixin.js
+++ b/src/mixins/raster-layer-line-mixin.js
@@ -104,7 +104,6 @@ export default function rasterLayerLineMixin(_layer) {
     return state
   }
 
-
   function doJoin() {
     return state.data.length > 1
   }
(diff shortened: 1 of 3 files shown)


src/mixins/raster-layer-line-mixin.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable complexity */
```
cjvil, line 1: we voted this rule out of Immerse long ago, in case the current generation is interested in doing the same for charting...

src/mixins/raster-layer-poly-mixin.js
```diff
@@ -320,6 +354,8 @@ export default function rasterLayerPolyMixin(_layer) {
           as: "color"
         })
       }
+      // }
```
cjvil, line 357: cleanup

PR #656, opened 2023-10-10
[FE-17174] Check to make sure a source with id doesn't already exist to avoid mapbox error

cmatzenbach:
If you zoom in the chart editor and cancel, you will occasionally run into a mapbox error saying "a source with this id already exists".  Add an additional check by source id to ensure the source doesn't already exist and avoid console error if it does.


johallar (approved):
LGTM! Little troubling that this ever worked? Maybe source and ID are the same a lot of the time?

PR #658, opened 2023-11-17
[FE-16380] Contour chart: Filtering values does not affect overall color range

cjvil:
Adds `auto-contour` color domain handling to calculate domain from min/max; cannot use existing "auto" domain logic as-is as that presumes there is a `color` column.

### DIFF
--- a/src/mixins/raster-layer-poly-mixin.js
+++ b/src/mixins/raster-layer-poly-mixin.js
@@ -442,6 +442,10 @@ export default function rasterLayerPolyMixin(_layer) {
     return state.encoding.color && state.encoding.color.domain === "auto"
   }
 
+  function usesAutoContourColors() {
+    return state.encoding?.color.domain === "auto-contour"
+  }
+
   function getPolygonScale({ state, layerFilter, layerName, autocolors }) {
     const useColorScale = !(state.encoding.color.type === "solid")
     let scale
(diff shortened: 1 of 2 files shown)


src/mixins/raster-layer-poly-mixin.js
```diff
@@ -644,13 +649,34 @@ export default function rasterLayerPolyMixin(_layer) {
       })
     }
 
+    // Contour polys may be colored by contour value rather than a color column
+    if (contourAutocolors) {
```
johallar, line 653: I think it would be preferable to roll this auto color logic into the if/else above `if(isContourType(state)) { ... } else { ... }` and use the same "auto" value if possible. That way we could avoid creating a special value for contour chart in `auto-contour`

cjvil: `auto` is used in the logic to pull scales for legends pretty removed from this context; I’d rather not add more exceptions/pass extra layer type context around if there is no overlapping logic with `auto`. If/when contour supports legends, sure

johallar: Makes sense but I wonder if there's another way to disable the legend without a special value, or if the decision could be made above this where a reference to chart is available (`toLegendState` or something). Either way, works for now until legends are a thing

johallar (approved):

PR #659, opened 2023-11-17
[FE-17337] Move sample to end of query for metis codegen

cmatzenbach:
This moved the `SAMPLE_RATIO` part of the query codegen in metis to the end of the `WHERE` clause.  See ticket for details.

We may need to fix this for contour as well, but need to consult with Dr. Flaxman.  This addresses the immediate SCE issue.

### DIFF
--- a/src/mixins/raster-layer-line-mixin.js
+++ b/src/mixins/raster-layer-line-mixin.js
@@ -237,6 +237,20 @@ export default function rasterLayerLineMixin(_layer) {
       })
     }
 
+    if (typeof filter === "string" && filter.length) {
+      transforms.push({
+        type: "filter",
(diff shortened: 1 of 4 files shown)


johallar (changes requested):
This is a great opportunity for a unit test that checks if the sample transform is last in the list to prevent regressions in the future. I think we should at least add a comment above them, unit test much preferred

tmostak:
Do we have any examples or tests that show changed behavior by moving SAMPLE_RATIO to the end? I'm not saying it wouldn't, but except for it manifesting a bug/deficiency in backend fragment skipping logic I don't see how this would change backend behavior.

johallar (approved):

PR #669, opened 2024-05-17
[FE-17481] Preven table sorting option from being hidden

cmatzenbach:

Fixes issue where if table chart was resized to a small size, the column alignment buttons would float over the column name and sort button.  Now the alignment buttons will move as close as they can to the column title before getting cut off.  Moved these buttons to their own wrapper, with a flexbox layout and less padding so the buttons can get closer to the column title before being cut off.

Closes [FE-17481](https://heavyai.atlassian.net/browse/FE-17481)

### :mag: Verification Steps

1. Follow repro steps in corresponding jira ticket
2. Ensure the alignment buttons no longer overlay over the sort button and you can still sort a column when the table is resized to be very small.

### :camera_flash: Screenshot

https://github.com/heavyai/heavyai-charting/assets/15268289/c68b0b1a-dd83-4507-a45d-a056fb8cebee

[FE-17481]: https://heavyai.atlassian.net/browse/FE-17481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dist/charting.css
+++ b/dist/charting.css
@@ -801,8 +801,6 @@ body {
     .dc-chart .table-header-item .left-align-btn, .dc-chart .table-header-item .center-align-btn, .dc-chart .table-header-item .right-align-btn {
       min-width: 18px;
       min-height: 24px;
-      right: 0;
-      top: 1px;
       padding: 4px 0px;
       cursor: pointer;
       visibility: hidden;
--- a/dist/charting.css
+++ b/dist/charting.css
@@ -790,22 +790,28 @@ body {
     position: relative; }
   .dc-chart .table-header-item {
     position: relative;
-    white-space: nowrap; }
+    white-space: nowrap;
+    display: flex;
+    align-items: center;
+    flex-direction: row;
+    border-bottom: 1px solid #e2e2e2; }
+    .dc-chart .table-header-item .align-btn-container {
+      display: flex;
+      align-items: center; }
     .dc-chart .table-header-item .left-align-btn, .dc-chart .table-header-item .center-align-btn, .dc-chart .table-header-item .right-align-btn {
-      width: 18px;
-      height: 24px;
-      position: absolute;
+      min-width: 18px;
+      min-height: 24px;
       right: 0;
       top: 1px;
       padding: 4px 0px;
       cursor: pointer;
-      display: none; }
+      visibility: hidden; }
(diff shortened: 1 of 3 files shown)


dist/charting.css
```diff
@@ -784,28 +784,35 @@ body {
...
+    .dc-chart .table-header-item .align-btn-container {
+      display: flex;
+      align-items: center; }
     .dc-chart .table-header-item .left-align-btn, .dc-chart .table-header-item .center-align-btn, .dc-chart .table-header-item .right-align-btn {
-      width: 18px;
-      height: 24px;
-      position: absolute;
+      min-width: 18px;
+      min-height: 24px;
       right: 0;
```
johallar, line 804: Don't need these positioning properties top/right if using flex to position

scss/chart.scss
```diff
@@ -676,16 +677,23 @@ body {
...
+          display: flex;
+          align-items: center;
+        }
         .left-align-btn, .center-align-btn, .right-align-btn {
-            width: 18px;
-            height: 24px;
-            position: absolute;
+            min-width: 18px;
+            min-height: 24px;
             right: 0;
```
johallar, line 691: whoops i mean here

mapd-bot-os:
build succeeded

scss/chart.scss
cmatzenbach: Good catch, fixed

mapd-bot-os:
build succeeded

johallar:
I noticed the text alignment controls can overflow and be hidden, we should probably make sure the min column width accommodates the sorting and the alignment controls:

Screen Shot 2024-05-20 at 8 48 35 AM

mapd-bot-os:
build succeeded

cmatzenbach:
> I noticed the text alignment controls can overflow and be hidden, we should probably make sure the min column width accommodates the sorting and the alignment controls:
> Screen Shot 2024-05-20 at 8 48 35 AM

@johallar Ok, after spending hours playing around with this, unfortunately we can't do it with CSS.  D3 is manually setting the size of the header and table cells on the elements when you re-size, and it has a set min-width it will only go to.  By setting a min-width on the header, the header becomes mis-aligned with the table columns.  We also can't decide on a standard min-width since the column names for the header vary in size.  I tried to make it respect the elements present in that container (i.e. the sort buttons, column names, and alignment buttons) but the only way I could get it close was a lot of hacky manual resizing with d3 code, nothing worked with css.  I'll open an UI improvement ticket that can be worked on later, as this at least solves the bug and offers an improvement over current UX.

johallar (approved):

PR #673, opened 2024-07-30
Add hash coloring functions for d3 charts; call when value is string

cmatzenbach:
- Add hash function into charting
- When assigning colors to string values, use hash function instead of d3 coloring

### DIFF
--- a/src/mixins/color-mixin.js
+++ b/src/mixins/color-mixin.js
@@ -213,11 +213,9 @@ export default function colorMixin(_chart) {
     if (typeof measureColor === "string") {
       const hash = cyrb53(measureColor)
       const colorIndex = hash % colors.length
-      console.log(colors[colorIndex])
       return colors[colorIndex]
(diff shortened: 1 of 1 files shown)


src/mixins/color-mixin.js
```diff
@@ -166,7 +166,11 @@ export default function colorMixin(_chart) {
     const range = _chart.colors().range()
     const middleColor = range[Math.floor(range.length / 2)]
 
-    return _colors(_colorAccessor.call(this, data, index)) || middleColor
+    const value = _colorAccessor.call(this, data, index)
+
+    return typeof value === "string"
```
cmatzenbach, line 171: Probably safe to do this for all data types, since hashing function handles numbers as well as strings, but just have it turned off for strings atm

mapd-bot-os:
build failed

johallar (approved):

PR #677, opened 2024-09-25
[FE-17951] D3 palette mappings

cmatzenbach:
See Immerse PR for details


src/charts/pie-chart.js
```diff
@@ -182,15 +181,9 @@ export default function pieChart(parent, chartGroup) {
...
-    if (_chart.customDomain().length > 0 && chartData) {
-      pieData = chartData.filter(d => _chart.customDomain().includes(d.key0))
-    } else if (chartData) {
-      pieData = chartData
-    }
-
     // if we have data...
-    if (pieData && d3.sum(pieData, _chart.valueAccessor())) {
-      pieData = pie(utils.maybeFormatInfinity(pieData))
+    if (chartData && d3.sum(chartData, _chart.valueAccessor())) {
```
cmatzenbach, line 185: Reverting this back to how it was before...do not want to filter data based off immerse-set domain

src/charts/pie-chart.js
```diff
@@ -140,13 +146,6 @@ export default function pieChart(parent, chartGroup) {
         })
     })
   })
-
-  override(_chart, "getColor", (data, index) => {
```
cmatzenbach, line 144: No longer need override since others are handled below

src/utils/color-helpers.js
```diff
@@ -0,0 +1,43 @@
...
+    ])
+  )
+  chart.customDomain([...dataMap.keys()])
+  chart.customRange([...dataMap.values()])
+}
+
+export function maybeUpdateAllOthers(chart, data, domain, range) {
+  if (
+    data.map(d => d.data.key0).includes(ALL_OTHERS_LABEL) &&
+    !domain.includes()
```
johallar, line 28: What's up with this `includes()` check?

src/utils/color-helpers.js
```diff
@@ -0,0 +1,43 @@
...
+  ) {
+    domain.push(ALL_OTHERS_LABEL)
+    range.push(ALL_OTHERS_COLOR)
+  } else if (
+    domain.includes(ALL_OTHERS_LABEL) &&
+    !data.map(d => d.data.key0).includes(ALL_OTHERS_LABEL)
```
johallar, line 34: Data is looped through twice in this function, would be good to extract to variable at the top of the function.

src/utils/color-helpers.js
```diff
@@ -0,0 +1,43 @@
...
+  data,
+  dataAccessor,
+  domain,
+  range,
+  wrapKey = false
+) {
+  const dataArr = data.map(dataAccessor)
+  const domainRangeMap = new Map([...domain].map((key, i) => [key, range[i]]))
+  const dataMap = new Map(
+    [...dataArr].map((key, i) => [
```
johallar, line 15: `[...domain]` and `[...dataArr]` not necessary, `map` returns a new array

src/utils/color-helpers.js
```diff
@@ -0,0 +1,43 @@
...
+      key,
+      domainRangeMap.get(key) ??
+        chart.getColor(wrapKey ? { key0: key } : key, i)
+    ])
+  )
+  chart.customDomain([...dataMap.keys()])
```
johallar, line 21: I don't think these spread operators are necessary either, `keys` and `values` return new arrays

src/utils/color-helpers.js
cmatzenbach: Bad auto find and replace lol

src/utils/color-helpers.js
cmatzenbach: Good point, extracted both of those to the top of the function

johallar (approved):

PR #680, opened 2024-11-13
[FE-18003] Bug fix - modifying mapping and cancelling removed mapping

cmatzenbach:
Issue was a fail to safeguard against `customRange` being undefined, which threw an error, but then also was triggering a `setCustomDefaultOtherColorValue` call in immerse, which wiped out the `paletteMappingId` on `chart.color` for D3 charts, causing the mapping not to be selected when returning to the chart.


johallar (approved):
Needs updated `dist` files, otherwise lookin good

PR #687, opened 2024-12-18
[FE-18087] Raster chart re-render reverts legend sorting

cjvil:
Ensures getLegendStateFromChart returns sorted domain + makes value of color.sorted match current sort direction


cmatzenbach (approved):
Works great!

PR #688, opened 2025-01-22
[FE-18151] Full color hashing

cmatzenbach:
See Immerse PR for details: https://github.com/heavyai/immerse/pull/9460

### DIFF
--- a/src/mixins/raster-layer-point-mixin.js
+++ b/src/mixins/raster-layer-point-mixin.js
@@ -285,12 +285,15 @@ export default function rasterLayerPointMixin(_layer) {
       ) {
         transforms.push({
           type: "project",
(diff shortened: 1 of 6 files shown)


src/mixins/raster-layer-point-mixin.js
```diff
@@ -527,6 +541,17 @@ export default function rasterLayerPointMixin(_layer) {
         }
       }
     } else {
+      parser.writeSQL({
```
johallar, line 544: Leftover stuff? Or does this have some side effects?

cmatzenbach: Good catch - that was some leftover stuff from experimenting

PR #696, opened 2025-02-13
[FE-18169] In fullColorHashing mode, add infinite scroll to legend

cmatzenbach:
Depends on legendables PR: https://github.com/heavyai/legendables/pull/12

When `fullColorHashing` is enabled, revise the legend so that it has infinite scrolling, so that it will continuously load up more values from crossfilter until it has reached the end of the data.  This also changes the initial rendering of colors on the legend - rather than pulling colors from the initial domain (as defined by immerse), it will instead first look for any customized colors, and hash the rest of them to match with the new `fullColorHashing` querying.

By default loads results in batches of 50.  Legend filtering should still work, and should only load as many columns as are available per the current bounding box.  Also moves hashing function to separate helper function.

### DIFF
--- a/src/chart-addons/stacked-legend.js
+++ b/src/chart-addons/stacked-legend.js
@@ -307,9 +307,9 @@ export async function getLegendStateFromChart(chart, useMap, selectedLayer) {
                     color.originalDomain,
                     color.originalRange,
                     color.defaultOtherRange,
-                    color?.fullColorHashing,
+                    color.fullColorHashing,
                     color.palette.val,
-                    color?.customColors
+                    color.customColors
                   )
                 : {}
 
--- a/src/chart-addons/stacked-legend.js
+++ b/src/chart-addons/stacked-legend.js
@@ -192,7 +192,7 @@ function getUpdatedDomainRange(
   }
 }
 
-const buildFullHashedDomainRange = (
+const dedupeAndAppendValuesToDomainRange = (
   currentDomain,
   currentRange,
   colValues,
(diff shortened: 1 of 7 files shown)


src/chart-addons/stacked-legend.js
```diff
@@ -115,27 +132,80 @@ async function getTopValues(layer, layerName, size) {
     return newCf
       .dimension(dimension)
       .group()
-      .topAsync(size, OFFSET)
+      .topAsync(
```
cmatzenbach, line 135: Needed to add additional `ORDER BY` clause to make query deterministic and ensure the same data doesn't get returned across multiple queries due to tiebrakers in the `val` column, so need to pass in extra default params

src/chart-addons/stacked-legend.js
```diff
@@ -105,8 +108,22 @@ function setColorScaleDomain_v1(domain) {
...
+  const chartId = layer?.crossfilter()?.chartId
+  const dimension = layer?.getState()?.encoding?.color?.field
+  const newCf = layer?.crossfilter()?.cloneWithChartId(chartId, layerName)
+
+  if (chartId && dimension && newCf) {
+    const allValues = await newCf
+      .dimension(dimension)
+      .group()
+      .all()
+      .then(res => res)
```
johallar, line 121: Remove `.then(res => res)`

src/chart-addons/stacked-legend.js
```diff
@@ -115,27 +132,80 @@ async function getTopValues(layer, layerName, size) {
...
-  const newDomainRange = new Map(
-    [...newDomain].map(key => [key, oldDomainRange.get(key) ?? defaultColor])
-  )
+  const newDomainRange = fullColorHashing
+    ? new Map(
+        [...newDomain].map(key => [
+          key,
+          customColorsDomainRange && customColorsDomainRange.get(key)
+            ? customColorsDomainRange.get(key)
+            : determineColorByValue(key, palette)
```
johallar, line 180: Code golf nit:
`customColorsDomainRange?.get?.(key) ?? determineColorByValue(key, palette)`

src/chart-addons/stacked-legend.js
```diff
@@ -238,7 +308,10 @@ export async function getLegendStateFromChart(chart, useMap, selectedLayer) {
                     colValues,
                     color.originalDomain,
                     color.originalRange,
-                    color.defaultOtherRange
+                    color.defaultOtherRange,
+                    color?.fullColorHashing,
```
johallar, line 312: Not sure if there's any utility in the optional chaining here, it will fail before this if color is nullish

cmatzenbach: Since `fullColorHashing` is a new prop added this release, this is to protect against legacy charts not having it in their `color` object

src/chart-addons/stacked-legend.js
cmatzenbach: Good one

johallar (approved):
See note in immerse about trying to avoid double scroll bars, but this is great!

src/chart-addons/stacked-legend.js
johallar: The optional chaining here is protecting against `color` being undefined or null. 

If the old charts don't have `fullColorHashing` in the `color` object, this will return `undefined` (as expected I assume) with or without the optional chaining as you have it here.

cmatzenbach: Oh yeah, nevermind, you're right, won't matter since we know at this point `color` is defined

PR #703, opened 2025-11-03
Upgrade node to 24.11.0

cmatzenbach:

1. Removes node-sass from the project
2. Bump internal dependencies to eliminate node-sass from dependency tree
3. Upgrade node version to latest LTS (24.11.0)
4. Use compatible webpack plugin
5. Upgrade `sass` and `sass-loader` to match immerse; upgraded `webpack` to v5 for `sass-loader`

Closes [HERC-133](https://jirasw.nvidia.com/browse/HERC-133)


package.json
```diff
@@ -115,21 +115,22 @@
...
     "lodash": "^4.17.21",
+    "mini-css-extract-plugin": "^0.12.0",
     "mocha": "^9.2.2",
     "mocha-jenkins-reporter": "^0.1.9",
-    "node-sass": "^4.14.1",
     "nyc": "^15.0.0",
     "prettier": "^1.19.1",
     "proxyquire": "1.7.4",
-    "sass-loader": "6.0.7",
+    "sass": "^1.93.2",
```
johallar: Any reason for a newer version of sass here than in immerse ("^1.77.0"), i'd think it would be best to be consistent across our repos if possible, if not 🤷  but we should if we can

cmatzenbach: Upgraded `sass` and `sass-loader` to match immerse - also needed to upgrade to webpack 5 for `sass-loader`

PR #706, opened 2026-05-27
Replace force-resolutions with overrides and update deps per security scan

cmatzenbach:
Previously since we were locked into node 14 we had to use the `npm-force-resolutions` package to force update transient dependencies, but now that we upgraded node that is no longer needed.  Thus, this removes that package, removes the `preinstall-if-root` script that ran it, and replaces the `resolutions` section with the native `overrides`. Also updates the following deps per the security scan:
- axios
- protocol-buffers-schema (transient dep)


johallar (approved):
Looks good, you'll need to include the built files as well before merging

PR #707, opened 2026-06-01
Add license headers, remove mapbox token from examples

cmatzenbach:
Adds required license headers to all source files for OSS release. Also removes raw mapbox tokens, which were placed into each example file. Replaced with the string `YOUR_MAPBOX_TOKEN`.


dist/charting.js.LICENSE.txt
johallar, line 1: Do you know what's up with this new file?

johallar (changes requested):
Interesting new license file generated for the compiled output, probably worth figuring out what this is about since it seems very relevant to adding licensing header to everything

dist/charting.js.LICENSE.txt
cmatzenbach: Did some digging, and it's auto-generated by `terser`, and contains license headers from 3rd party libraries in a project. It's only supposed to contain licenses of 3rd party libraries, not the main library, so it doesn't need to be modified to include our headers.

cmatzenbach: Seems like a dependency update bumped `terser` to a version that had this extraction behavior enabled by default. I had actually seen it before on my dependency updates branch but never pushed it, as I wasn't sure what it was and was trying to get things in before Pey ran another scan. But looking into it now, it should be included and is correct as-is.

johallar (approved):

PR #711, opened 2026-06-22
Add third party licenses file and generator script 

cmatzenbach:
Adds `THIRD_PARTY_LICENSES.md` file, generator script, and updated README to explain how to run script to update.


package.json
```diff
@@ -114,6 +114,7 @@
     "expose-loader": "^5.0.0",
     "http-server": "^0.11.2",
     "jsdom": "^11.7.0",
+    "license-checker": "^25.0.1",
```
johallar: We should be able to remove this dep as well, since it's held in the new repo right?

johallar (approved):

